PR #112, opened 2015-07-13
Fix Splunk Streaming Command dict_reader.py attempts to iterate over none

### LINKED ISSUE #110: Splunk Streaming Command dict_reader.py attempts to iterate over none
Hey guys,

For some reason Splunk trys to iterate over none and throws a `TypeError: 'NoneType' object is not iterable` in https://github.com/splunk/splunk-sdk-python/blob/master/splunklib/searchcommands/splunk_csv/dict_reader.py#L38. This is only reproducible in in certain environments it seems (happens in our production distributed cluster(s) but not in local test VMs). This causes the command to still execute but throws an error on the screen cause Splunk sent it no data.

I understand that it _shouldn't_ do this yet still throws the error even though it enriches the data correctly. We were able to duplicate this occurrence with both a custom command and with the `countmatches` example code in our production testbed

kkirsche:
Close #110

Fix provided by @David-Noble-at-work

### DIFF
--- a/splunklib/searchcommands/splunk_csv/dict_reader.py
+++ b/splunklib/searchcommands/splunk_csv/dict_reader.py
@@ -31,7 +31,7 @@ class DictReader(csv.DictReader, object):
             try:
                 self._fieldnames = self.reader.next()
             except StopIteration:
-                pass
+                self._fieldnames = []
             self.line_num = self.reader.line_num
             self.__mv_fieldnames = []
             self.__fieldnames = []


shakeelmohamed:
Thanks for doing this work @kkirsche, can you please make a PR against the develop branch?

kkirsche:
@shakeelmohamed @David-Noble-at-work  seems you have the fix in Develop and have since August 11, 2014. Any reason you don't have it in your master branch then....?

shakeelmohamed:
@kkirsche Ah, thanks for pointing that out. We will have a new release out the door shortly :)

kkirsche:
No worries @shakeelmohamed. It's just very concerning in regards to development process that it's been there a year and it isn't in your master branch. Your master has newer files than develop which signals to me a red flag that it isn't being used fully as a develop branch :\

cc: @David-Noble-at-work @glennblock

itay:
@kkirsche for some reason, that commit from Nov 24 is literally the only one missing from master that is in develop from that timeframe. For some reason it looks like it didn't get merged over when we did a release in March. We will get to the bottom of why that happened, but apologies for the problem it caused. 

We do use develop exclusively for development, and no changes are made to master directly that are not documentation changes like the contribution guidelines, etc.

kkirsche:
@itay I'm not so sure I can believe you with regards to that anymore — You also don't have your format function for your Set validator from 9 months ago when your last release was after that. These types of lapses in master that exist in develop so long ago raise large concerns for me.

cc: @glennblock

itay:
@kkirsche I did get to the bottom off it. When Python 2.7.9 came out, we did a hotfix release that just updated support for it. It did not include any of the fixes we had pending in `develop`, which is why you're seeing a case where there are some fixes in `develop` that are not in `master`, but `master` also has some "newer" commits.

You are right that we haven't done a new full Python SDK release in quite a while - we were focused on a few other things. However, we're prepping one now that should be out in the next few weeks.

kkirsche:
Thank makes sense. I guess my frustration is that in the current state I consider the SDK broken and in need of a hot fix to bring it to where it was supposed to be even if it's not a feature release. I understand the focus elsewhere it just leaves me wondering where the SDK lies if you don't plan to bring master to parity until "a few weeks"

PR #121, opened 2015-10-16
Make splunklin.binding.ResponseReader.readline function.

cjw296:
I bumped into this trying to do:

```
>>> conn = client.connect(...)
>>> result = conn.jobs.export('search ...', output_mode='csv')
>>> result.readline()
'"'
```

The problem is the 1 byte peek followed by a one byte read here:

https://hg.python.org/cpython/file/v2.7.5/Modules/_io/iobase.c#l466

This results in an attempt to read 0 bytes from the HTTPResponse in result._response, which ends up hitting this code path:

https://hg.python.org/cpython/file/v2.7.5/Lib/httplib.py#l567

We ask for 0 bytes, and sure enough, we get 0 bytes, but httplib sees this as "no data left" and closes the connection.

I couldn't figure out how best to test this, or indeed, how to run the test suite in such a way that the existing tests on HEAD of master all passed before I started work. But, functional testing has shown this to work.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -1270,7 +1270,8 @@ class ResponseReader(io.RawIOBase):
         self._buffer = ''
         if size is not None:
             size -= len(r)
-        r = r + self._response.read(size)
+        if size != 0:
+            r = r + self._response.read(size)
         return r
 
     def readable(self):


itay:
@cjw296 sorry for the delayed response. Could you:

1. Sign the contribution agreement?
2. Rebase this on top of develop (and change the target to be develop), so that you pick up the Travis CI tests?

cjw296:
Given this software is so poor, I'm unlikely to bother with a wet-ink contributor agreement.
You can do you own rebasing.
Take it or leave it...

PR #130, opened 2015-12-10
Add better handling for Keep-Alive connections

itay:
Python's httplib generally has bad support for Keep-Alive connections,
which is why we never used it. We didn't explicitly ask for "Connection: Close",
but it was implicit. Additionally, splunkd would always return it.

However, an issue arises when talking to a Load Balancer which issues
a Keep-Alive request to splunkd, and returns that header to us. Python
then goes into a particular path which the code didn't handle gracefully.

What this change does is two-fold:

1. Always request to close the connection via the "Connection: Close" header.

2. If we get back a "Connection: Keep-Alive" header, we properly handle
it by not immediately closing our reference to the connection and only
doing so when the response has been completed.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -1235,8 +1235,9 @@ class ResponseReader(io.RawIOBase):
     # For testing, you can use a StringIO as the argument to
     # ``ResponseReader`` instead of an ``httplib.HTTPResponse``. It
     # will work equally well.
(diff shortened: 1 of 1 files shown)


David-Noble-at-work:
LGTM

Qmando:
This fixes the problem I was having. Thanks.

PR #132, opened 2015-12-23
Addition of basic auth support

matthewsullivan-wf:
Following the pattern defined in https://github.com/splunk/splunk-sdk-ruby/pull/43, I have implemented Basic Auth in this PR.

```
from splunklib.client import Service

service = Service(host="splunk.myhost.net", port=8089, scheme="https", version=6.3,
                  basic=True, username="steve", password="puppydoge")

service.indexes['test'].submit("My message!", host="localhost", sourcetype="text")
```

PR #133, opened 2016-01-29
Add Basic Auth support

itay:
This change was contributed by matthewsullivan-wf, and adds Basic Auth
support to the Python SDK. You can now pass a username and password when
creating the Service, and then also pass a `basic=True` argument to tell
it to use Basic Auth.

This supercedes #132

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -33,6 +33,7 @@ import io
 import sys
 import Cookie
 
+from base64 import b64encode
 from datetime import datetime
 from functools import wraps
 from StringIO import StringIO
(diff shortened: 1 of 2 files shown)


matthewsullivan-wf:
Thanks @itay!  If we can get a release cut in the next few weeks I'd super appreciate it.  We're excited to get our org running :)

matthewsullivan-wf:
@itay - when might there be a new release cut that includes this change?  We have several systems that we'll be plugging into our Splunk install and the dev teams are asking me about it.  Thanks!

PR #140, opened 2016-04-07
Removed all obscured URLs

loganknecht:
Removed obscured URLs. Also fixed a broken link in the splunklib/searchcommands/search_command.py.


glennblock:
Thanks @loganknecht!

PR #146, opened 2016-07-06
Fix 'maxresultrows' extraction bug

likel:
Get the value of maxresultrows from searchinfo object in metadata instead of metadata. The metadata object has no attribute named 'maxresultrows'.


shakeelmohamed:
Thanks @likel, the changes look good!

PR #147, opened 2016-07-26
Identical entity names will cause an infinite loop

darlingtld:
Identical entity names will cause an infinite loop "RuntimeError: maximum recursion depth exceeded". Give a clear message about which applications have this entity.
Trying to `self.name` in `Raise AmbiguousReferenceException` will result in a dead loop. 
```
splunklib\client.py", line 928, in _load_atom_entry
    raise AmbiguousReferenceException("Fetch from server returned multiple entries for name %s." % self.name)
  splunklib\client.py", line 1067, in name
    return self.state.title
  splunklib\client.py", line 1091, in state
    if self._state is None: self.refresh()
  splunklib\client.py", line 1011, in refresh
    self._state = self.read(self.get())
  splunklib\client.py", line 1071, in read
    results = self._load_state(response)
splunklib\client.py", line 934, in _load_state
    entry = self._load_atom_entry(response)
client.py", line 928, in _load_atom_entry
    raise AmbiguousReferenceException("Fetch from server returned multiple entries for name %s." % self.name)
client.py", line 1067, in name
    return self.state.title
.
.
```

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -933,7 +933,10 @@ class Entity(Endpoint):
     def _load_atom_entry(self, response):
         elem = _load_atom(response, XNAME_ENTRY)
         if isinstance(elem, list):
-            raise AmbiguousReferenceException("Fetch from server returned multiple entries for name %s." % self.name)
+            apps = [ele.entry.content.get('eai:appName') for ele in elem]
+
+            raise AmbiguousReferenceException(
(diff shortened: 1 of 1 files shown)


itay:
@darlingtld thanks for the fix, a couple of things:

1. Can you sign the CLA?
2. Can you re-target this at develop?
3. Do you have a sample EAI XML blob that causes this? Could we add it as a test?

darlingtld:
@itay 
I have signed the CLA and I attached a sample json in the commit for testing.

PR #149, opened 2016-07-30
"handlers" stanza missing in examples/searchcommands_template/default/logging.conf

doksu:
Please see diff.


doksu:
I think there's a problem with the automated travis-ci checks. When using the searchcommands_template app on Splunk 6.4.1 the diff I've put in this pull request is valid and required to fix a bug.

PR #159, opened 2016-10-28
index.submit() doesn't work with a namespaced service

adrianj98:
A name space is not needed for indexing


shakeelmohamed:
Thanks @adrianj98, we came up with a simpler fix 089841a49578196f0dbdadc039999c83a3748641. It will be out in the next release

PR #160, opened 2016-10-28
DVPL-7026 added "Splunk " to the authentication of streaming sockets …

adrianj98:
…if using a token


shakeelmohamed:
Thanks @adrianj98  we just merged the fix with some test coverage, it will be out in the next release

PR #162, opened 2016-11-02
remove Inputs.kind check to reduce performance impact on each rest call

zhengyuli:
Hi, we encountered seriously performance issue when using splunklib.Inputs object, we found the root cause is splunklib.Inputs object will check the `kind` for each rest call, during check process splunklib will trigger lots of additional rest calls if there are too many kinds of data inputs in Splunk.

For the detailed info please go to: https://confluence.splunk.com/display/PROD/ERD+Performance+and+Design+Deficiency+in+SplunkLib+Inputs

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -2430,15 +2430,12 @@ class Inputs(Collection):
         :return: The relative endpoint path.
         :rtype: ``string``
         """
-        if kind in self.kinds:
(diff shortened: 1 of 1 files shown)


wjo1212:
dears, would you please merge the fixing? Thanks!

shakeelmohamed:
Replaced by #166

Issue #91, opened 2014-08-05
Support Python 3


tylereaves:
Currently splunklib only suppports Python 2. This is quite a bit annoying for those of us that have made the jump to Python 3,  especially as the spunk api is not easily handleable via standard Python XML tools.

filmor:
Would you be open to adding `six` as a dependency?

davidbono:
+1

rhaarm:
I started writing a Python 2&3 compatible version with Six and future, but after making changes the Test Cases were not working and many revisions were coming in from the Splunk developers. I might revisit the Python 3 compatibility when time allots again.

itay:
@filmor we've been talking about it internally. The main issue we have with requiring `six` (or any other module) is that it would break the current pattern of not having any external dependencies. This is doubly important for us because of how some people use the SDK in their apps - they simply copy the `splunklib` folder into their app and don't need to worry about any dependencies.

If there was a way for us to not break that (i.e. not require pip/easy_install, have `splunklib` be entirely self contained), we'd be open to it.

Ghosttsix:
From the Six documentation on https://pythonhosted.org/six/

Six provides simple utilities for wrapping over differences between Python 2 and Python 3. It is intended to support codebases that work on both Python 2 and 3 without modification. **six consists of only one Python file, so it is painless to copy into a project.** (emphasis mine)

Is it possible just to keep your SDK updated with the latest version of six prepackaged?

dmikov:
+1 it's time guys

shakeelmohamed:
> Is it possible just to keep your SDK updated with the latest version of six prepackaged?

Since `six` is MIT licensed, I think this is reasonable especially since the license and version number are embedded in `six.py`.

Submitting a PR should be much easier now that we have support for Travis CI, so you don't have to run tests locally.

wjo1212:
cannot agree more, considering pack six as a part of Public SDK (in its sub-folder like package/six.py), refer to request way

tltx:
Pack six with the code and get it done. This is one of the last blockers for a large project I'am working on to move over to Python 3. It's time!

swedishmike:
+1 from me as well. I'm working on a specific project where I really don't want to have to drop down to 2.7 just to get the data into Splunk using the SDK.

buysse:
[...]

PR #171, opened 2017-05-04
Decode binary response

highfestiva:
The old Python 2 code concatenates the http-received data with a string,
which no longer works as Python 3 receives http data in binary form.

Simple decoding might be naive, as it might break non-utf8 or (or possibly
end up in the middle of a codepoint, not sure), but it works afaict.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -62,6 +62,8 @@ DEFAULT_HOST = "localhost"
 DEFAULT_PORT = "8089"
 DEFAULT_SCHEME = "https"
 
+http2str = (lambda s:s) if sys.version_info < (3,) else (lambda b:b.decode())
+
 def _log_duration(f):
     @wraps(f)
     def new_f(*args, **kwargs):
@@ -1289,7 +1291,7 @@ class ResponseReader(io.RawIOBase):
         self._buffer = ''
         if size is not None:
             size -= len(r)
-        r = r + self._response.read(size)
+        r = r + http2str(self._response.read(size))
         return r
 
     def readable(self):


shakeelmohamed:
@highfestiva thanks. Will this change be backwards compatible? In order for us to support Python 3, we must also be backwards compatible with at least Python 2.7

highfestiva:
@shakeelmohamed aah, didn't know that. I want to come back with a fix, but what is the best way to do it in this case where py2 carry string type and py3 bytes? Since this vital part of the code is not working, perhaps you have no "standard" way of handling it?

highfestiva:
My changes are now py2-compliant, but a lot the present code is not it seems. The state of the py3 Travis test builds seem less-than optimal, to put it prudently?

shakeelmohamed:
@highfestiva Yes, most of the SDK is not Python 3 compatible. There's quite a bit of work to get the entire SDK to a fully 2 and 3 compatible state

highfestiva:
Why would we want to be py2-compatible at all? van Rossum's point, flawed or not, of skipping backwards compatibility, was that part of the syntax in py2 was inconsistent and/or poor. Ubuntu's already over on py3, and the splunklib api doesn't change. I suspect more newcomers would have problems with py2 than py3 versions, and since the splunklib api is unchanged I can't see why we would want to hang on to 2.7 at all?

shakeelmohamed:
@highfestiva please see the comments in #91. Dropping Python 2 support is non-negotiable for us at this time

highfestiva:
The stated reason for keeping py2 in 2014 was "given the platform support matrix we have as well as some libraries still not being available on Python 3." @shakeelmohamed: would you mind elaborating on that?

shakeelmohamed:
@highfestiva Splunk needs to continue supporting (and shipping) Python 2 because there are several Splunk-developed Python libraries written in Python 2 that are shipped with Splunk. Dropping Python 2 support as a company is not feasible because Splunk needs to continue supporting legacy operating systems (that ship Python 2 as the default).

Does that make sense?

rhaarm:
For a product that annoyingly claims to be "an enterprise solution like no
other and would take a lot of effort with open-source software" seems to be
very resistant to the Python 3 support when all of those open-source
solutions have it out of box.

highfestiva:
@shakeelmohamed: sort of I guess. In that 1980'ies IBM way, maybe. At some point in time you are going to transition. One way to do it sooner would be to build and endorse a py3 version while sticking a sunset date on the old goo. That would make more sense to many (most?), and make the progressive company profile you're aspiring to more belivable. I love and promote your product with a passion, but this I find slightly annoying and counter-productive.

PR #173, opened 2017-05-26
Fix Configuration(distributed=True) resulting in a command that is NOT distributed

datasearchninja:
As per support Case #488821, DVPL-7118, DEVC-1

In a generating command using SCP2 (chunked = true), commands are unexpectedly distributed

#!/usr/bin/env python

import sys
import time
import socket

from splunklib.searchcommands import dispatch, GeneratingCommand, Configuration, Option, validators

@Configuration(
  distributed=False
)
class distfalse(GeneratingCommand):
    def generate(self):
      yield {'_time': time.time(), '_raw': 'This ran on ' + socket.gethostname()}
      self.flush()

dispatch(distfalse, sys.argv, sys.stdin, sys.stdout, __name__)

### DIFF
--- a/tests/searchcommands/test_configuration_settings.py
+++ b/tests/searchcommands/test_configuration_settings.py
@@ -81,7 +81,7 @@ class TestConfigurationSettings(TestCase):
 
         self.assertIs(command.configuration.distributed, False)
         self.assertIs(command.configuration.generating, True)
-        self.assertEqual(command.configuration.type, 'streaming')
+        self.assertEqual(command.configuration.type, 'stateful')
 
         command.configuration.distributed = True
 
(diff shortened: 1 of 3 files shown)


tests/searchcommands/test_configuration_settings.py
```diff
@@ -81,7 +81,7 @@ def generate(self):
```
dan1: I had to change this to

 [('generating', True), ('type', 'stateful')])

to get the test to pass.

tests/searchcommands/test_configuration_settings.py
```diff
@@ -81,7 +81,7 @@ def generate(self):
 
         self.assertIs(command.configuration.distributed, False)
         self.assertIs(command.configuration.generating, True)
-        self.assertEqual(command.configuration.type, 'streaming')
+        self.assertEqual(command.configuration.type, 'stateful')
```
dan1: I had to set this back to get the test to pass.
self.assertEqual(command.configuration.type, 'streaming')

dan1:
disttrue_no_fix
distfalse_no_fix
disttrue_fix
distfalse_fix

dan1:
disttrue_no_fix

dan1:
disttrue_fix

PR #176, opened 2017-06-29
Add Python 3 support

tltx:
Here is some work I done to add Python 3 support. It's not finished and I can't get the tests to work in python 2 so it's hard to know what more needs to be done. If you have any tips on how fix "Connection refused" problem with the tests I can test more myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ language: python
 python:
   - "2.7"
   - "2.6"
+  - "3.5"
 
 install: "pip install unittest2"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -558,3 +558,4 @@ API reference under the section on accessing Splunk resources at:
 ## 0.1.0 (preview)
 
 * Initial Python SDK release
+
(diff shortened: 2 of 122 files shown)


tltx:
I have fixed the remaining issues I found. The test suite is kind if flakey but that does not seem to be related to Python 3. I think this PR is ready for regression testing and review.

tdhellmann:
Thanks so much for this PR! We're definitely eager to take a look, and our intention is to do what we can to take the change. That said, we also want to be upfront that we're in the middle of the next release of Splunk Enterprise at the moment, so it's going to take us a little bit of time to review the PR and land it. We want to make sure we're ready on our end to fully support Python 3 after the PR is merged, and that requires some dedicated time on our end.

Thanks again for the contribution!

brandonmburroughs:
@tdhellmann Is there any update here?  I've been using the Splunk Python SDK and would love for it to have Python 3 support and better integrate with the rest of my project!  I'd be happy to help @tltx update whatever conflicts there now are if this will get tested/merged.  Thanks!

PR #177, opened 2017-06-30
add six.py

tltx:
A good start to support Python 3 is to add a bundled six.py. It has a lot of useful stuff for supporting both Python 2 and Python 3 at the same time and it's the de facto standard way of doing this.


tltx:
The failed check has nothing to do with the PR as far as I can see

PR #180, opened 2017-07-17
Bugfix dvpl-7207

ljiang1:


itay:
@ljiang1 can you provide an explanation of what the bug is and what the fix is? I can't figure it out from the changes.

ljiang1:
The issue is that SDK didn't reply correctly to the getinfo request following the chunk protocol and just go ahead do execute action. This used to work before 6.4 because there is only one getinfo call anyway. But since 6.5 which the new optimizer code start to make  multiple getinfo calls from different threads and if they didn't get reply from the external process, they will not terminate the thread and SDK will just go ahead do multiple execute. This pr add the reply msg to splunk so that the non-execute getinfo thread knows they need to terminate.

splunklib/searchcommands/search_command.py
```diff
@@ -758,7 +776,6 @@ def _process_protocol_v2(self, argv, ifile, ofile):
             self.finish()
             exit(1)
 
-        self._record_writer.write_metadata(self._configuration)
```
shakeelmohamed: Why did we remove this?

splunklib/searchcommands/generating_command.py
```diff
@@ -174,6 +174,11 @@ class SomeCommand(GeneratingCommand)
         streaming = false
 
     """
+
+    # overwrite reply_message_v2
+
+    reply_message_v2=(('type', "streaming"),("generating",True))
```
leverich: I'm not sure I follow this. How would I make a GeneratingCommand that is stateful (i.e. runs on the search head) or reporting (i.e. produces results in the Statistics tab and not the Events tab in the S&R UI?)?

splunklib/searchcommands/search_command.py
```diff
@@ -683,6 +686,21 @@ def _process_protocol_v2(self, argv, ifile, ofile):
         # noinspection PyBroadException
         try:
             self._record_writer = RecordWriterV2(ofile, getattr(self._metadata.searchinfo, 'maxresultrows', None))
+
+            # fix dvpl-7207: send reply to splunk and wait for reply of execute action
```
leverich: There's something troubling I find about this comment, I'm just not sure that I can see any logic issues in the code yet.

Basically, the concern is this: it assumes that the SDK sends a chunk to splunkd and splunkd REPLIES to that chunk with something else.

However, splunkd should be the one initiating every action. splunkd sends the SDK a chunk (which is, effectively, a command to go do something), and the SDK replies with a response. In all cases, the SDK should be waiting for that command (e.g. either "getinfo" or "execute") from splunkd before doing anything. That is, there must be a 1:1 relationship between every chunk splunkd sends and every chunk the SDK replies with.

If there's ever an inversion of that notion (e.g. if the SDK is sending a chunk and waiting for a response from splunkd), it is very likely to lead to a protocol violation from the two getting out of lock-step.

ljiang1:
@leverich  Thanks! I will start a new pull request based on the suggestion.

ljiang1:
open new pr based on Jacob's comments

PR #181, opened 2017-07-23
Generatorfix

ljiang1:
check if the action is execute before execute the generate command


tests/searchcommands/test_searchcommands_app.py
```diff
@@ -355,7 +355,7 @@ def _get_search_command_path(self, name):
     def _load_chunks(self, ifile):
         import re
 
-        pattern = re.compile(r'chunked 1.0,(?P<metadata_length>\d+),(?P<body_length>\d+)\n')
+        pattern = re.compile(r'chunked 1.0,(?P<metadata_length>\d+),(?P<body_length>\d+)')
```
shakeelmohamed: Should we make the newline optional at the end with `\n?`

tests/searchcommands/test_searchcommands_app.py
```diff
@@ -368,6 +368,8 @@ def _load_chunks(self, ifile):
                 break
 
             match = pattern.match(line)
+            if match is None:
+                continue
```
shakeelmohamed: If we have this logic, the following `assertIsNotNone` is not needed

itay:
@ljiang1 I'm still kind of confused how this fixes the issue?

ljiang1:
@itay As Jacob suggested in previous pr, comparing to streamingcommand and other commands, before they call the customer supplied execution, they executed the searchcommands._records_protocol_v2(), which read the reply from splunk and check if the reply action is execute; but the generatingcommand didn't call _records_protocol_v2(). the fix is adding the same checking code as _records_protocol_v2(). The reason can't do the same calling format as streamingcommand is that the way of _records_protocol_v2() implements prevents the yielding result, so I end up just add the check code here.

ljiang1:
@leverich @itay  by doing the read_chunk of splunk reply, somehow it terminates the getinfo only processes that added by optimizer code... maybe it is defined by the protocol? @leverich

leverich:
If the script is being invoked just for getinfo, splunkd will close the processes' stdin after it receives the getinfo reply, so the SDK should read an EOF. Since this change adds the _read_chunk to _execute, it will detect the EOF and return None and return early from _execute. Thus, the GeneratingCommand's generate() method is never called in the case that this is just an invocation for getinfo.

I believe this change should address the immediate bug we're seeing (generate() method being invoked multiple times when it shouldn't be).

I think there's a separate issue with the GeneratingCommand class in that it only ever emits one chunk (e.g. you couldn't have a generating command that generates an infinite stream of events). That's a pre-existing problem and doesn't need to be addressed as part of this bugfix.

tests/searchcommands/test_searchcommands_app.py
```diff
@@ -377,8 +379,8 @@ def _load_chunks(self, ifile):
             body_length = int(match.group('body_length'))
             body = ifile.read(body_length) if body_length > 0 else ''
 
-            if len(chunks) == 0:
-                self.assertEqual(ifile.readline(), '\n')  # the getinfo exchange protocol requires this
+            # if len(chunks) == 0:
```
leverich: How come this is commented out? I'd prefer to not see new commented-out code committed to the repo.

tests/searchcommands/test_searchcommands_app.py
ljiang1: resolved.

splunklib/searchcommands/search_command.py
```diff
@@ -873,8 +873,10 @@ def _read_chunk(ifile):
         # if body_length <= 0:
         #     return metadata, ''
 
+        body=""
         try:
-            body = ifile.read(body_length)
+            if body_length>0:
```
shakeelmohamed: nit: spaces

splunklib/searchcommands/search_command.py
```diff
@@ -873,8 +873,10 @@ def _read_chunk(ifile):
         # if body_length <= 0:
         #     return metadata, ''
 
+        body=""
```
shakeelmohamed: nit: spaces

PR #182, opened 2017-08-14
Fix Configuration(distributed=True) resulting in a command that is NOT distributed

shakeelmohamed:
Replaces #173


shakeelmohamed:
@dan1 to fix the tests

shakeelmohamed:
LGTM, waiting on a second review from @itay or @ljiang1

Issue #192, opened 2018-02-28
Loading data from JSON throws error when values are unicode


ruxandraburtica:
## Problem:
I've created a Splunk job using the general Splunk examples and I'm waiting for the data (continually checking to see if the job finished):

```
        job = jobs.create(search, **search_kwargs)
        print('Generated job with SID: {}\n\n'.format(job.sid))

        while True:
            while not job.is_ready():
                pass
            stats = {'isDone': job['isDone'],
                     'doneProgress': float(job['doneProgress']) * 100,
                     'scanCount': int(job['scanCount']),
                     'eventCount': int(job['eventCount']),
                     'resultCount': int(job['resultCount'])}

            status = ('\r%(doneProgress)03.1f%%   %(scanCount)d scanned   '
                      '%(eventCount)d matched   %(resultCount)d results') % stats

            sys.stdout.write(status)
            sys.stdout.flush()
            if stats['isDone'] == '1':
                sys.stdout.write('\n\nDone!\n\n')
                break
            sleep(2)

        results_iterator = job.results(count=200)
        raw_data = results_iterator.read()
        entire_json_data = splunk_data.load(raw_data)
```

## Details:

*0. Version of this project:*
`splunk-sdk==1.6.2`

*1. Platform version:*
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.3 LTS
Release:	16.04
Codename:	xenial
```

*2. Framework version:*
`Python 2.7.12`

*3. Splunk version:*
`Splunk 6.5.3.1`

*4. Stacktrace:* 
```
  File "/work/data_processing/connectors/splunk.py", line 112, in get_and_wait_for_data
    entire_json_data = splunk_data.load(raw_data)
  File "/work/venv/lib/python2.7/site-packages/splunklib/data.py", line 83, in load
    return load_root(items[0], nametable)
  File "/work/venv/lib/python2.7/site-packages/splunklib/data.py", line 145, in load_root
    k, v = load_elem(element, nametable)
  File "/work/venv/lib/python2.7/site-packages/splunklib/data.py", line 109, in load_elem
    value = load_value(element, nametable)
  File "/work/venv/lib/python2.7/site-packages/splunklib/data.py", line 172, in load_value
    name, item = load_elem(child, nametable)
  File "/work/venv/lib/python2.7/site-packages/splunklib/data.py", line 109, in load_elem
    value = load_value(element, nametable)
[...]

PR #193, opened 2018-02-28
Using basestring instead of str so that unicode values are also accepted.

ruxandraburtica:
PR for https://github.com/splunk/splunk-sdk-python/issues/192

### DIFF
--- a/splunklib/data.py
+++ b/splunklib/data.py
@@ -112,7 +112,7 @@ def load_elem(element, nametable=None):
     if attrs is None: return name, value
     if value is None: return name, attrs
     # If value is simple, merge into attrs dict using special key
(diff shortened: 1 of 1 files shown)


shakeelmohamed:
Hi @ruxandraburtica, this change is not compatible with Python 3 and has runtime errors on CI since `basestring` is not defined.
Please use `string_types` after adding `from six import string_types` at the top of the file.

ruxandraburtica:
@shakeelmohamed thanks it's great to see that `python3` support is on its way 👍 
I have fixed the issue, I've used the already imported six module, please let me know if that is ok

shakeelmohamed:
Looks great thank you!

PR #200, opened 2018-04-13
Lower threshold for SSL verification to Python 2.7.5

treydock:
This avoids errors with SSL verification on Python version shipped with RHEL 7


mew1033:
Maybe with the ability to turn on/off SSL verification with a parameter introduced in https://github.com/splunk/splunk-sdk-python/commit/5807d9076e708f21f3949fafea304485396b5d99, this check should be removed. If the server is running a valid certificate, there's no reason to disable verification unless it's needed. Now people will just be able to disable it where they need it.

dan1:
This should resolve the problem with being able disable SSL verify.  https://github.com/splunk/splunk-sdk-python/pull/233

PR #206, opened 2018-06-06
Fix XML responses to not throw errors for unicode characters

loganknecht:
Changed `splunklib/data.py` `load` function to convert text provided into a utf-8 string so that Python XML parser will not throw errors when processing utf-8 encoded responses.

### DIFF
--- a/splunklib/data.py
+++ b/splunklib/data.py
@@ -76,7 +76,7 @@ def load(text, match=None):
         'namespaces': [],
         'names': {}
     }
-    root = XML(text)
+    root = XML(text.encode('utf-8'))
     items = [root] if match is None else root.findall(match)
     count = len(items)
     if count == 0: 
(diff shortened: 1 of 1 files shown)


splunklib/data.py
```diff
@@ -76,6 +77,12 @@ def load(text, match=None):
         'namespaces': [],
         'names': {}
     }
+
+    # Convert to unicode encoding in only python 2 for xml parser
+    if sys.version_info < (3, 0, 0):
```
shakeelmohamed: we can use `and` and combine these into a single `if`

splunklib/data.py
```diff
@@ -79,9 +79,8 @@ def load(text, match=None):
     }
 
     # Convert to unicode encoding in only python 2 for xml parser
-    if sys.version_info < (3, 0, 0):
-        if isinstance(text, unicode):
-            text = text.encode('utf-8')
+    if(sys.version_info < (3, 0, 0) and isinstance(text, unicode)):
```
shakeelmohamed: nit: outer parens aren't necessary but LGTM

Issue #207, opened 2018-06-12
400: Bad Request for POST Requests to Change Permissions


ianpittwood:
# Issue Description
Using splunklib.client.post to change permissions on eventtypes fails because the client strips out the owner and sharing fields from the body of the request. This results in the body only containing perms.read and perms.write without specified owner or sharing fields. Since these fields are missing, the request returns a 400 Bad Request error code from the Splunk server.

## Instance Information
* splunk-sdk version: 1.6.4
* Platform version: macOS 10.13.5
* Framework version: Python 3.6.5
* Splunk version: 7.0.3

## Example Code Snippet
```
# Where self is a wrapper for the Splunklib client
# o = self.service.username
# sharing = 'user'
# perms.read = '*'
# perms.write = '*'
path = '/' + '/'.join([SPLUNK_REST_NAMESPACE, self.service.username, self.app, SPLUNK_REST_EVENT_TYPES, name, SPLUNK_REST_PERMISSION])
body_dict = {
            'owner': owner,
            'sharing': sharing,
            'perms.read': perm_read,
            'perms.write': perm_write,
}
self.service.post(path, **body_dict, output_mode='json')
```

## Correct REST API Request
`curl -k -u <user>:<password> <splunk_address>/servicesNS/<user>/<app>/saved/eventtypes/<name>/acl -d owner=<user> -d sharing=user -d perms.read=* -d perms.write=*`

## Stack Trace
File "/opt/project/splunk_utils/event.py", line 312, in chmod_event_type
    return self.service.post(path, **body_dict, output_mode='json')
  File "/usr/local/lib/python3.6/site-packages/splunklib/binding.py", line 289, in wrapper
    return request_fun(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/splunklib/binding.py", line 71, in new_f
    val = f(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/splunklib/binding.py", line 742, in post
    response = self.http.post(path, all_headers, **query)
  File "/usr/local/lib/python3.6/site-packages/splunklib/binding.py", line 1208, in post
    return self.request(url, message)
  File "/usr/local/lib/python3.6/site-packages/splunklib/binding.py", line 1228, in request
    raise HTTPError(response)
splunklib.binding.HTTPError: HTTP 400 Bad Request -- b'{"messages":[{"type":"ERROR","text":"The following required arguments are missing: owner, sharing."}]}'

[...]

PR #208, opened 2018-06-12
Fix HTTP 400 errors for certain post requests

ianpittwood:
This is one possible fix I came up with for issue #207. By moving owner, app, and sharing from arguments to kwargs in the post method, they will be included in the request body. This will fix any 400 errors that result from owner, app, or sharing being left out of REST requests that require them to be explicitly defined in the body of the request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,28 +282,16 @@ If you would like to contribute to the SDK, go here for more information:
 
 ### Support
 
-1. You will be granted support if you or your company are already covered 
-   under an existing maintenance/support agreement. Send an email to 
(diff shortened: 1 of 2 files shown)


tdhellmann:
@sketchypotato - thanks for the (very detailed!) issue report and PR!

One quick note - this approach will cause a breaking change, so we're going to take a look at doing this without changing the method signature.

tdhellmann:
See also: https://github.com/splunk/splunk-sdk-python/issues/231

ashah-splunk:
@ianpittwood thank you for the PR, the issue #207 has been fixed sometime back hence closing this PR

PR #210, opened 2018-06-19
Add an await_completion method on Job just to simplify the common case

cheddar:
This PR creates an `await_completion" method on the Job class just to make the common case of "I submitted a search and now just want to wait until the results are ready, without copy-and-pasting a while loop around everywhere" more simple-like.

I tried to figure out how to add a test for it as well, but I think my general lack of familiarity with Python mixed with lack of familiarity with this code made it hard for me to figure out how to add the test.  As such, I haven't actually run this code yet, so there could be glaringly obvious problems and I apologize.  I decided to generate the PR to ask for help in creating the test.  

Can you provide any pointers for how I can test this?  I was basically looking for a test that was waiting on results, but didn't really find one...

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -2887,6 +2887,38 @@ class Job(Entity):
         self.post("control", action="unpause")
         return self
 
+    def await_completion(self, sleep_interval, callback=lambda done, progress: False):
+        """Waits for the job to complete, checking every sleep_interval seconds.
+        If callback is provided, will call the callback for every poll interval
+        with an indication of completion and progress.  If the callback returns
+        true, then the method will return, giving control back to the caller.
+
+        Regardless of the return value of the callback, await_completion always 
+        returns control back to the caller once the job has completed.
+
+        :param `sleep_interval`: The number of seconds between each poll
+        :type value: ``integer``
+        :param `callback`: Optional. A function of two arguments, the first is a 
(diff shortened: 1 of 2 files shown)


splunklib/client.py
```diff
@@ -2887,6 +2887,38 @@ def unpause(self):
...
+        :param `callback`: Optional. A function of two arguments, the first is a 
+        boolean representing if the job has completed.  The second is a floating 
+        point number representing how much of the job has completed.  Returns a 
+        boolean that, if True, causes the await_completion call to return, if 
+        False, polling will continue.
+        :type value: ``function``
+
+        :return: The :class:`Job`
+        """
+        while True:
```
shakeelmohamed: This can be re-written without a `while True` loop.

splunklib/client.py
```diff
@@ -2887,6 +2887,38 @@ def unpause(self):
         self.post("control", action="unpause")
         return self
 
+    def await_completion(self, sleep_interval, callback=lambda done, progress: False):
```
shakeelmohamed: @cheddar how do you plan to call this function? We haven't use callbacks as a pattern in the SDK, and don't recommend we do so here.
Also the word `await` implies some async behavior which is not preset, so I would suggest renaming this function to `wait_for_completion()`

PR #221, opened 2018-09-17
Update the explorer example to Python 3

Photon89:


tdhellmann:
Hey @Photon89 - we're trying to maintain compatibility with both Python 2 and Python 3 for the time being, and it looks like this approach will cause issues with Python 2. Could you look into using the equivalent methods in six?

Photon89:
I fear, I can't - I don't even know Python, just wanted to use the API explorer and quickly researched how to make it work with Python 3. By the way, as far as I remember, I couldn't use the API explorer with Python 2 either but I'd need to check again.

shakeelmohamed:
Replaced by #224

PR #227, opened 2018-10-23
Use self._connection

kimox2:
Fix bug when close


tdhellmann:
Hey @kimox2  - thanks for the PR and thanks for pointing this out! We'll need to figure out what's causing the build failures before we can move on this, though - looks to be specific to Python 2.7.

shakeelmohamed:
Thanks for the PR @kimox2, the CI failure is an intermittent one not caused by your change. I've re-run CI and all builds have passed.

PR #235, opened 2019-01-06
Possible fix for #207?

wcastello:
Could anyone please check if this would be ok for fixing #207 ? 

Uses __ for service.post() on binding.py for arguments like app, owner, sharing needed for example for post requests to the /acl path for changing permissions and also adds an access_update() to client.py so that the client can transparently call an update in permissions without having to use the __ in the start of those arguments. 

At least I believe this approach won't cause a breaking change by changing the method signature as said by @tdhellmann on https://github.com/splunk/splunk-sdk-python/pull/208

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -746,7 +746,7 @@ class Context(object):
         if headers is None:
             headers = []
 
-        query = { key.lstrip('__'): val for key, val in query.items() }
+        query = { key.lstrip('_dup__'): val for key, val in query.items() }
 
         path = self.authority + self._abspath(path_segment, owner=owner, app=app, sharing=sharing)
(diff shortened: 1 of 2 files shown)


wcastello:
Hum, this is breaking now because there are calls like post(__conf=name), would have to think of another prefix or something.

wcastello:
Ok I'll look into the tests better later to see what's happening but it still seems to be due to the prefix addition/strip.

fantavlik:
I believe this has been addressed in #352 and now released as https://github.com/splunk/splunk-sdk-python/releases/tag/1.6.15 please upgrade via `pip install splunk-sdk==1.6.15` and let us know if you continue to see these issues.

PR #236, opened 2019-01-08
Don't output close tags if you haven't written a start tag

scottsavarese:
In looking at Splunk I found lots of events with just `</stream>` as the text. Looking at the code, the event_write class will always output the closing tag even if it doesn't output a start tag or any event. So, I added a check to see if there was a start tag written. If so, put a valid close tag. Otherwise, do not


scottsavarese:
I should have done this in develop. My apologies

PR #237, opened 2019-01-08
Don't output close tags if you haven't written a start tag

scottsavarese:
In looking at Splunk I found lots of events with just `</stream>` as the text. Looking at the code, the event_write class will always output the closing tag even if it doesn't output a start tag or any event. So, I added a check to see if there was a start tag written. If so, put a valid close tag. Otherwise, do not


bparmar-splunk:
Hi @scottsavarese,
Will you please re-push your changes (w. r. t. [newer sdk change](https://github.com/splunk/splunk-sdk-python/releases/tag/1.6.18))?
 
When your changes are pushed, it will trigger Github workflow for executing test cases. So that It will help us to troubleshoot if any issue is encountered.

scottsavarese:
@bparmar-splunk , Wow, old change. Forgot about this one. What exactly do you need me to do? Are you asking me to create a new PR for this that is more current, or simply pull the latest from the develop branch and fix any conflicts?

scottsavarese:
@bparmar-splunk , I pulled in the latest from your develop branch and fixed a merge conflict. The workflow won't start without a maintainer to kick it off. Let me know if you need me to do anything else.

bparmar-splunk:
Hi @scottsavarese,
We have approved workflow to execute test cases.
Test are passed in python v2, but fails in v3.
Cause of failure is due to the line which you have modified. It should be string instead of bytes w. r. t. python v3. (Please check error logs of Github CI workflow executed against python v3).

Could you please make relevant change so that it should be compatible with either of the version ?

scottsavarese:
@bparmar-splunk , I removed the bytes part of the output, but still seeing failures. They look 100% unrelated to my change. But I have no idea how to read your test results output.

PR #238, opened 2019-01-10
Move splunklib to lib folder instead of bin folder

xuchen509:
Putting splunklib in an app's /bin directory is highly discouraged, this change moves splunklib to apps' lib folder instead in searchcommands_app example


shakeelmohamed:
I just merged master -> develop, and develop -> update-splunklib-dep so the diff is more legible

examples/searchcommands_app/package/lib/splunklib
```diff
@@ -0,0 +1 @@
+/Users/xcheng/repo/github/splunk-sdk-python/splunklib
```
shakeelmohamed: this probably shouldn't be here

examples/searchcommands_app/package/bin/generatehello.py
```diff
@@ -17,11 +17,13 @@
 
 from __future__ import absolute_import, division, print_function, unicode_literals
 import app
-
-from splunklib.searchcommands import dispatch, GeneratingCommand, Configuration, Option, validators
 import sys
 import time
-from six.moves import range
+
+from splunk.clilib.bundle_paths import make_splunkhome_path
```
shakeelmohamed: It's not great that we have to import from the `splunk` package (aka the Python lib that ships in core) because our tests now become a bit more complicated.

How intense is `make_splunkhome_path`? Can we bring that into the Python SDK?

shakeelmohamed: Also this is causing some test failures as expected `ImportError: No module named splunk.clilib.bundle_paths` - since we're not running the tests using the Splunk Python. Now, this gets interesting when we look at Python 3 support because I don't know if that function is written with support for both... probably isn't. So if we bring the function into the SDK it needs to be written using `six` for dual compatibility

examples/searchcommands_app/package/lib/splunklib
xuchen509: done

examples/searchcommands_app/package/bin/generatehello.py
xuchen509: it's pretty hard to pull `make_splunkhome_path` to Python SDK by itself as it uses some other functions in `splunk` package. 

This is an example app, so it'll have to be run within Splunk anyway and we don't have to worry about Python 3 as long as we are using the stuff Splunk python supports. I'm not familiar with the tests we have here, are we not testing the app within a Splunk instance? how are we testing these custom search commands actually working?

shakeelmohamed: Yeah if you look at the tests, they actually just run under the context of the shell Python. 
It's probably possible to run those tests using something like `$SPLUNK_HOME/bin/splunk cmd python file.py` so we can give that a try

As for the Python 3 issue, looks like we're just skipping search commands tests if not running on Python 2.7 so don't worry about that. See `tests/searchcommands/__init__.py`

examples/searchcommands_app/package/bin/sum.py
```diff
@@ -19,8 +19,11 @@
 import app
 import sys
 
-from splunk.clilib.bundle_paths import make_splunkhome_path
-sys.path.append(make_splunkhome_path(['etc', 'apps', 'searchcommands_app', 'lib']))
+try:
```
shakeelmohamed: Can we lift all of this stuff to `__init__.py` in this directory?

PR #241, opened 2019-01-31
fix tests on windows

xuchen509:
Support running tests on windows for searchcommands_app example.


tests/searchcommands/test_searchcommands_app.py
```diff
@@ -416,7 +416,7 @@ def _run_command(self, name, action=None, phase=None, protocol=2):
                             ofile.write(b)
                 with io.open(uncompressed_file, 'rb') as ifile:
                     env = os.environ.copy()
-                    env['PYTHONPATH'] = ":".join(sys.path)
+                    env['PYTHONPATH'] = os.pathsep.join(sys.path)
```
xuchen509: thanks @lowell80 for good catch

PR #245, opened 2019-02-14
Fix the docs

lowell80:
I found and corrected a number of minor typos and syntax issues in the docs (as hosted at https://splunk-python-sdk.readthedocs.io/en/master/).  On this branch Sphinx can now build the docs without any errors.  (Specifically 01558e4 fixes the exceptions occurring during the build, which I suspect is what was causing RTD to says that the docs are "failing")

I split the changes into smaller commits so if there's objections to changes they'd be easier to isolate.


shakeelmohamed:
Thanks so much @lowell80 this is an incredible amount of work! We'll keep poking at CI and hope to get this merged shortly

shakeelmohamed:
Thanks so much for this work @lowell80!

PR #247, opened 2019-02-23
Performance improvement for updating or deleting input

### LINKED ISSUE #168: Performance issue to delete/update an input
If I want to delete one data input, I have to list all of inputs, filter them out, and delete one. It has big performance impact when there are many data inputs in my Splunk ENV, even if I already know the data input type & name.

shakeelmohamed:
Closes #168

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -463,6 +463,13 @@ class Service(_BaseService):
         response = self.get("/services/server/info")
         return _filter_content(_load_atom(response, MATCH_ENTRY_CONTENT))
 
+    def input(self, name, kind=None):
+        """Retrieves an input by name, and optionally type.
+
+        :return: A :class:`Input` object.
+        """
+        return Input(self, name, kind=kind).refresh()
+
     @property
     def inputs(self):
         """Returns the collection of inputs configured on this Splunk instance.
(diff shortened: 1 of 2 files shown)


tests/test_input.py
```diff
@@ -16,8 +16,6 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
-import unittest2
-
```
loganknecht: Good catch

splunklib/client.py
```diff
@@ -463,6 +463,13 @@ def info(self):
         response = self.get("/services/server/info")
         return _filter_content(_load_atom(response, MATCH_ENTRY_CONTENT))
 
+    def input(self, path, kind=None):
```
loganknecht: Not a big deal, but should we follow Go style practices of adding a new prefix to the declaration?

Like this
`def new_input(self, path, kind=None):`.

splunklib/client.py
```diff
@@ -463,6 +463,13 @@ def info(self):
...
+        """Retrieves an input by path, and optionally kind.
+
+        :return: A :class:`Input` object.
+        """
+        return Input(self, path, kind=kind).refresh()
```
loganknecht: **Non-blocking Personal Opinion**: I like to see these operations broken down atomically because it makes things easier to debug.

```
new_input = Input(self, path, kind=kind)
new_input.refresh()
return new_input
```

splunklib/client.py
shakeelmohamed: not if we want consistency, so I'm going to say let's not do that here

splunklib/client.py
shakeelmohamed: fair, I'm just following the style I used for `job()`

PR #250, opened 2019-03-11
SDK unicode issue for multi value fields.

docyes:
This is a fix for supporting multi value unicode fields.


vmalaviya-splunk:
As of now, we are not accepting any changes for Python versions lower than 3
As this change is not compatible with Python3 and later, We are closing the PR for now.

PR #262, opened 2019-07-24
properly add parameters to request based on the method of the request

### LINKED ISSUE #261: Possible Bug with binding class request method related to GET parameters
I created a lookup on `/servicesNS/-/-/data/lookup-table-files` to obtain the lookup table files. I noticed that the result was limited to 30 results per request. With that in mind, I attempted to pass in parameters for `output_mode=json&count=0`. The following code has been modified slightly and is based on a snippet found in the `invoke` function of `spurl.py` SDK example:

```
return binding.connect(**kwargs).request(path, method=method, body=urllib.urlencode({'output_mode': 'json', 'count': 0}))
```

The `output_mode` works properly and returns `json`, however when using `count` (or `offset`), it results in a 400 error:

```
HTTP 400 Bad Request -- {"messages":[{"type":"ERROR","text":"Argument \"count\" is not supported by this handler."}]}
```

Now, if I use `binding.connect(**kwargs).get` and pass in the same dictionary `{'output_mode': 'json', 'count' 0}` it gets properly passed as part of the GET request and returns all of the records.

```
return binding.connect(**kwargs).get(path, **{'output_mode': 'json', 'count': 0})
```

I believe the desired behavior on the `request` method would be to allow the use of Pagination and filtering parameters.

billmurrin:
This PR resolves #261 related to improper handling of `body` parameters within the `Context.request` method. The method used (`GET` or `POST`) will determine if the body is encoded as part of the url or passed as body into the request.

GET method requests (with and without a body) were tested and worked as expected.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -754,7 +754,7 @@ class Context(object):
 
     @_authentication
     @_log_duration
-    def request(self, path_segment, method="GET", headers=None, body="",
+    def request(self, path_segment, method="GET", headers=None, body={},
                 owner=None, app=None, sharing=None):
         """Issues an arbitrary HTTP request to the REST path segment.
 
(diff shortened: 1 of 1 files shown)


splunklib/binding.py
```diff
@@ -814,13 +814,27 @@ def request(self, path_segment, method="GET", headers=None, body="",
         path = self.authority \
             + self._abspath(path_segment, owner=owner,
                             app=app, sharing=sharing)
+
         all_headers = headers + self.additional_headers + self._auth_headers
         logging.debug("%s request to %s (headers: %s, body: %s)",
                       method, path, str(all_headers), repr(body))
-        response = self.http.request(path,
+
+        if body:
```
shakeelmohamed: Let's clean this up so we're only making on call to `self.http.request` and only modifying the dict we pass in

billmurrin:
Hi @shakeelmohamed. I've implemented the requested changes in the latest commit. Thanks!

billmurrin:
Hi @shakeelmohamed - the requested changes were made. For some reason, the CI issue is still ongoing.

shakeelmohamed:
@billmurrin This is not a high priority issue for us to investigate as we are severely under resourced at the moment. Unless this is 10/10 we will not look at it this week. Thank you for your patience & understanding.

cc @tdhellmann

KaterinaKop:
Any news here? We are looking forward to get this fix too!

PR #264, opened 2019-07-31
Disable SSL validation when verify=false

moshekaplan:
Previously, verify=false would only disable SSL validation if the user was running Python 2.7.9 or later. This commit allows users running Python versions before 2.7.9 to also disable SSL validation.


mew1033:
@moshekaplan I made another PR (#283) that fixes this.

shakeelmohamed:
Replaced by #283

PR #272, opened 2019-09-11
Parse is required because of line

mehmetcodes:
Try running this, it doesn't work.  Not only do you need parse, you need to handle the cmd


mehmetcodes:
Noted if you chmod +x part of the parsing issue goes away, but you still need the import in the example code

ashah-splunk:
@mehmetcodes thank you for the PR. We tried running the example without the Parse and it seems to be working fine in the latest SDK version, hence closing this PR. Please let us know if this issue still exists in the latest SDK

PR #273, opened 2019-09-12
HTTPError fix for Py2-Py3 (Issue 271)

Hetang23:
Resolving the errors as mentioned in the Issue-271 for _splunklib/binding.py_ and _splunklib/modularinput/event_writer.py_

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -1004,9 +1004,9 @@ class HTTPError(Exception):
     def __init__(self, response, _message=None):
         status = response.status
         reason = response.reason
-        if sys.version_info < (3, 0):
+        try:
             body = response.body.read()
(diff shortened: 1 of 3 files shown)


ljiang1:
@Hetang23 The tests in CI failed and look related to the changes. Can you help check your fix again?

Hetang23:
@ljiang1 yes, I'll be looking into it.

ljiang1:
@Hetang23 I have some concerns on the changes on event_writer.py, which cause the test failures. If you can provide some examples on the "document" objects for the changes you have made on line 85 (self._out.write(ET.tostring(document, encoding="unicode"))), we can evaluate more on if this change is only for specific cases  or when this should be used etc.

Issue #275, opened 2019-09-17
`ReportingCommand` not compatible with python 3


willbowditch:
I'm using a virtual env (similar to how [Splunk Scientific Computing app](https://splunkbase.splunk.com/app/2882/) does) to develop Python 3 applications whilst Splunk is still running Python 2.

The `splunk-sdk-python` says: 

> The Splunk SDK for Python requires Python 2.7+, including Python 3. The Splunk SDK for Python has been tested with Python v2.7 and v3.5.

However, I don't think this is true for `ReportingCommand` search commands. Any reporting command with a `def map():` function with `@Configuration()` decorator gives the following error:

```py
f = vars(command)[b'map']   # Function backing the map method
KeyError: b'map'
```

Which references this line:

https://github.com/splunk/splunk-sdk-python/blob/07278fafb1bef8b65638c2662d8962ecdf35326f/splunklib/searchcommands/reporting_command.py#L256

## Tests

Testing shows that the `b'map'` byte string indexing is not compatible with Python 3.

```py
class Test():
    def map():
        pass

vars(Test)[b'map']

print('Done!')
```

### Python 3

```sh
➜ python --version & python decorator_test.py

Python 3.5.7

Traceback (most recent call last):
  File "decorator_test.py", line 5, in <module>
    vars(Test)[b'map']
KeyError: b'map'
```

### Python 2

```
➜ python --version & python decorator_test.py
Python 2.7.9
Done!
```

shakeelmohamed:
Thanks for the bug report @willbowditch, we're open to a PR but currently preparing for .conf and don't have the bandwidth to investigate this further.

willbowditch:
@shakeelmohamed submitted a PR https://github.com/splunk/splunk-sdk-python/pull/276

Not familiar with the splunk-sdk but the patch seems to fix the example code here on my development build (Splunk 7.0.1): https://github.com/splunk/splunk-sdk-python/blob/master/examples/searchcommands_app/package/bin/sum.py

amysutedja:
Fixed and closed per https://github.com/splunk/splunk-sdk-python/pull/276

PR #276, opened 2019-09-18
Handle `b''` strings in Python 3

willbowditch:
Adds logic to fix bug described https://github.com/splunk/splunk-sdk-python/issues/275 where byte strings `b''` do not work for indexing or concatenation in Python 3. This causes `ReportingCommands` to fail if they have a `map()` func.

### DIFF
--- a/splunklib/searchcommands/reporting_command.py
+++ b/splunklib/searchcommands/reporting_command.py
@@ -24,6 +24,7 @@ from .streaming_command import StreamingCommand
 from .search_command import SearchCommand
 from .validators import Set
 from splunklib import six
+import sys
 
 
 class ReportingCommand(SearchCommand):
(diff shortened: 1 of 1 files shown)


willbowditch:
The failed travis CI is unrelated to the code changes. c7f9a0c and 038138f are identical apart from a dummy file used to trigger the build.

hanswurscht:
I'm hitting the same bug.
Can someone merge this please?

amysutedja:
After some investigation: I believe that we have fixed this issue as a side effect of fixing some other Python 3-related problems in 1.6.13. This code seems to be covered in `tests/searchcommands/test_searchcommands_app.py::TestSearchCommandsApp::test_sum_as_unit` which used to fail when run in Python 3, but now works.

Closing this PR.

PR #280, opened 2019-09-20
relax timeout to 3000

AmosChenYQ:


AmosChenYQ:
Close PR,because CI test succeed after relax timeout to 3000

PR #286, opened 2019-10-10
Implement Circle Builds with tox

rfaircloth-splunk:
This PR implements CircleCI builds with tox and Splunk 7.2, The current travis builds also implement 7.0 however this version is EOL. A followup enhancement should add 7.1 and 7.3 to the matrix. 

A number of tests are sensitive to timing or make use of client port binding which is not a good practice. Those tests continue to fail with this PR. This is not yet ready to replace travis but can be incremented upon by the community

Project admin will need to enable the circle CI project on the Splunk account


fantavlik:
We've since moved to Github Actions

PR #291, opened 2019-11-05
Remove clear_password assertions for storage/passwords tests

dgao1029:
POST requests to the /storage/passwords endpoint should not return the clear_password back in plaintext. The POST response should mask the clear_password but the Python SDK tests specifically check for the created or updated password. I removed those assertion checks.

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -1844,7 +1844,7 @@ class StoragePasswords(Collection):
 
         entries = _load_atom_entries(response)
         state = _parse_atom_entry(entries[0])
-        storage_password = StoragePassword(self.service, self._entity_path(state), state=state, skip_refresh=True)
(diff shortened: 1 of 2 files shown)


fantavlik:
Hey @dgao1029 I think the change that we actually want that will ensure that the clear_password is available as it was before to end-users is to force a state refresh after password creation. Following the Java SDK which you were just investigating this will update our `StoragePassword` model using the GET response.

So if you revert your testing changes and change this line: https://github.com/splunk/splunk-sdk-python/blob/develop/splunklib/client.py#L1847

to
```python
storage_password = StoragePassword(self.service, self._entity_path(state))
```

I believe that will allow users to retrieve the `password.clear_password` as they were before without removing functionality.

ashah-splunk:
@dgao1029 thank you for the PR, closing the PR as the change is not required.

Issue #274, opened 2019-09-17
EventWriter is writing bytes on stdout which is not supported in python3


chinmay2197:
Python `sys.stdout.write()` does not supporting writing bytes to stdout stream
In event_writer.py `self._out.write(b"<stream>")` at line 54 and `self._out.write(b"</stream>")` at line 81 
In event_writer.py `write_xml_document` method try to write encoded xml document to stream with `self._out.write(ET.tostring(document))`
In event.py `self._out.write(ET.tostring(document))`  in `write_to` method

This causes ERROR in modular input ingesting event like below:
`TypeError: write() argument must be str, not bytes`

shakeelmohamed:
Thanks for the bug report @chinmay2197, we're open to a PR but currently preparing for .conf and don't have the bandwidth to investigate this further.

croncar77:
I have the same error in write_xml_document method of event_writer.py file.

It seems to work changing 
self._out.write(ET.tostring(document)) to self._out.write(ET.tostring(document, encoding='unicode')) in event_writer.py file

schneewe:
You can also use this change (works for me):
self._out.write(ET.tostring(document)) to self._out.write(ET.tostring(document).decode())

rhymefororange:
Same for me. The write() function in python3 expects `str` and not `bytes`. And when I want to write and event to Splunk it just throws TypeError.
Are there any clues on when this may be fixed?

ByamB4:
Same for me. Working on python 2

ncanumalla-splunk:
Related to #271

ashah-splunk:
Latest release is compatible with Python3, so we would request you to try out with the latest version of the python sdk.

In the latest python sdk release,  before writing the encoded XML document to stream it is parsed through '[ensure_str](https://github.com/splunk/splunk -sdk-python/blob/13f07cd08f8b2017c1cdafc2fbc75673013dc713/splunklib/six.py#L901)' method  which handles the decoding of the bytes to string before encoding the xml document .

ashah-splunk:
Latest release of python sdk supports Python3 along with Python2, so we would request you to use the latest version of python sdk

In the latest version, before writing the encoded XML document to stream, it is encoded through "[ensure_str()](https://github.com/splunk/splunk-sdk-python/blob/13f07cd08f8b2017c1cdafc2fbc75673013dc713/splunklib/six.py#L901)" which handles the decoding of bytes before encoding the xml document while using Python3 

[...]

PR #295, opened 2019-11-26
EventWriter is writing bytes on stdout which is not supported in pyth…

chinmay2197:
…on3 #274

Resolved bug #274 for event_writer.py and event.py for modular input

### DIFF
--- a/splunklib/modularinput/event.py
+++ b/splunklib/modularinput/event.py
@@ -104,5 +104,5 @@ class Event(object):
         if self.done:
             ET.SubElement(event, "done")
 
-        stream.write(ET.tostring(event))
+        stream.write(ET.tostring(event).decode())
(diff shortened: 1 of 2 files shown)


tdhellmann:
Hi @chinmay2197 - thanks for the PR! Can you please fill out the contribution agreement located here: https://www.splunk.com/en_us/form/contributions.html before we proceed?

ashah-splunk:
@chinmay2197  Thank you for the PR.
Latest release of python sdk supports Python3 along with Python2, so we would request you to use the latest version of python sdk

In the latest version, before writing the encoded XML document to stream, it is encoded through "ensure_str()" which handles the decoding of bytes before encoding the xml document while using Python3

Do let us know if the issue is resolved .

PR #296, opened 2019-12-02
Send additional headers with Delete Requests.

Oz4P4X3:
Send additional headers with a delete request. Currently only GET, POST and REQUEST allow this.


ashah-splunk:
@Oz4P4X3 thank you for the PR. We would request you to also add a test case for the change suggested in the PR.

Ickerday:

PR #308, opened 2020-02-25
fix  AttributeError: 'module' object has no attribute '_create_unverified_context'

fengxx:
v1.6.12 contains PR #283 which breaks on legacy python without PEP 476 support


ashah-splunk:
Hi @fengxx , thank you for the PR. Splunk recommends using the newer Python (i.e version 3) . Also in near future Splunk will be removing its support for Python version 2, hence we would suggest you to use the newer Py3 with the latest Splunk SDK version. Please let us know if this is still an issue after upgrading to Python 3.

PR #314, opened 2020-04-01
Smoke tests

amysutedja:
This PR accomplishes the following.

## Moves toward use of Tox and Pytest

The existing use of `unittest` structures is antiquated, and modern Pythonic code uses `pytest` and `tox`. This PR introduces Tox and Pytest but does not (yet) integrate them into CI.

## Brings in missing tests and makes them Python 3.x-compatible

Because the existing `unittest` code had some custom integrations, they did not properly run tests in subdirectories. When running with `pytest`, these test results became visible, and it became apparent that they were not compatible with Python 3. This PR fixes those compatibility issues.

## Stricter .conf file syntax

The existing `.conf` files use inline comments. This is invalid because the `.ini` file format requires that comments and non-comments can't be defined on the same line. Python 2 was lenient in parsing these, but Python 3 is not. This PR makes all example `.conf` files valid.

## Marks important tests as "smoke"

It's helpful to have a set of tests that covers essential functionality in a short amount of time. Tests that exercise the following have been marked as `smoke` using the `pytest` marking facility:

* Authentication
* Search
* Modular Inputs
* Search Command

## Fixes minor issues in SearchCommands

The fixed Python 3 tests revealed minor issues in the SearchCommands module. Here's what's fixed:

* `Code` validator would crash on invalid code due to accessing the `.message` property, which is unavailable in Python 3.
* `EventingCommand`, `GeneratingCommand`, and `StreamingCommand` now properly appear as Python 3-style iterables. (This may change in the future, as the treatment of these commands as `dict` may be a bug.)

## Adds a script to build a .splunkrc

You might be interested in building a `.splunkrc` from some kind of JSON structure that describes a Splunk deployment. A new script (in the new `scripts` directory) nicely produces that file.


tests/searchcommands/test_internals_v2.py
```diff
@@ -257,76 +263,6 @@ def test_record_writer_with_random_data(self, save_recording=False):
...
-
-            # Write observed data (as an aid to diagnostics)
-
-            observed_path = expected_path + '.observed'
-            observed_value = ifile.getvalue()
-
-            with io.open(observed_path, 'wb') as ifile:
-                ifile.write(observed_value)
-
-            self._compare_chunks(observed, expected)
```
fantavlik, line 328: I see from your commit msg that this was removed due to portability issues, I'm curious to know more though not a blocker for the PR.

splunklib/searchcommands/eventing_command.py
```diff
@@ -139,4 +139,6 @@ def iteritems(self):
             iteritems = SearchCommand.ConfigurationSettings.iteritems(self)
             return imap(lambda name_value: (name_value[0], 'events' if name_value[0] == 'type' else name_value[1]), iteritems)
 
+        items = iteritems
```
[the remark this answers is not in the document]
amysutedja: Ouch, interesting. You're right that this is to handle `six.iteritems()`; in fact one of the commands already implemented it like this but three commands didn't, so my change pushes that out across the three commands to make them consistent.

I think we have a couple of options here.

1. Just leave `items` undefined in PY2, as was originally designed.
2. Throw a `NotImplementedError` in PY2.
3. Implement `items` correctly in PY2.

I assume in all cases we'd leave `iteritems` working in PY3. Thoughts?

tests/searchcommands/test_internals_v2.py
```diff
@@ -59,6 +61,9 @@
 
 max_length = 1 * 1024
 
+# generate only non-wide Unicode characters, as in Python 2, to prevent surrogate values
+MAX_NARROW_UNICODE = 0xD800 - 1
```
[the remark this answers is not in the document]
amysutedja: The ticket has been filed!

tests/searchcommands/test_internals_v2.py
amysutedja: So here's the basic problem.

> We have this test function called test_record_writer_with_recordings: https://github.com/splunk/splunk-sdk-python/blob/master/tests/searchcommands/test_internals_v2.py#L274
> 
> It appears to me that this is essentially test_record_writer_with_random_data except with static test fixtures: https://github.com/splunk/splunk-sdk-python/blob/master/tests/searchcommands/test_internals_v2.py#L174
> 
> Why I think this: I loaded it up in Python 2.x and it ends up building a dictionary with a fieldnames value of` OrderedDict([(u'fieldnames', [u'_serial', u'_time', u'random_bytes', u'random_dict', u'random_integers', u'random_unicode'])` which implies to me that it’s similar data as in the random_data version of the test.
> 
> Why do I say “in Python 2.x”? Well.
> 
> It seems as if the “recordings” are pickled data, which I take to mean that it’s an object that’s been serialized into a file.
> 
> The problem is, when you serialize internal object structures there’s no guarantee that they’re portable across versions.
> 
> So of course if you try to un-pickle in Python 3, the data — which I assume was originally serialized in Python 2 — is invalid.
>
> BUT the recordings seem to just be the same kind of stuff you’d get in random_data, just less random.
>
> But you could just reproduce that by feeding in the same Python seed. So I kind of feel like what if we just delete this function entirely

After working with @shakeelmohamed we came to the conclusion that this test scenario is fundamentally redundant anyway (as it can be easily regenerated via passing in the correct seed), so I cut it.

tests/searchcommands/test_decorators.py
```diff
@@ -391,38 +400,85 @@ def test_option(self):
...
+        #     "('required_fieldname', u'some.field_name'),"
+        #     "('required_file', u" + six.text_type(repr(__file__)) + "),"
+        #     "('required_integer', u'100'),"
+        #     "('required_map', 'foo'),"
+        #     "('required_match', u'123-45-6789'),"
+        #     "('required_optionname', u'some_option_name'),"
+        #     "('required_regularexpression', u'\\\\s+'),"
+        #     "('required_set', u'bar'),"
+        #     "('set', u'bar'),"
+        #     "('show_configuration', u'f')])")
```
ljiang1, line 431: remove these commented-out

tests/searchcommands/test_internals_v1.py
```diff
@@ -225,53 +232,56 @@ def test_input_header(self):
 
         input_header = InputHeader()
 
-        with closing(StringIO('\r\n'.encode())) as input_file:
+        print("String: %s" % type('\r\n'))
+        print("Encode: %s" % type('\r\n'.encode()))
+
```
ljiang1, line 237: just for debuging? remove these two lines

tests/searchcommands/test_search_command.py
```diff
@@ -690,15 +692,21 @@ def test_process_scpv2(self):
...
+        if six.PY2:
+            inspector = \
+                r'\"inspector\":\{\"messages\":\[\[\"ERROR\",\"StandardError at \\\".+\\\", line \d+ : test ' \
+                r'logging_configuration=\\\".+\\\" logging_level=\\\"WARNING\\\" record=\\\"f\\\" ' \
+                r'required_option_1=\\\"value_1\\\" required_option_2=\\\"value_2\\\" show_configuration=\\\"f\\\"\"\]\]\}'
+        else:
+            inspector = \
+                r'\"inspector\":\{\"messages\":\[\[\"ERROR\",\"Exception at \\\".+\\\", line \d+ : test ' \
+                r'logging_configuration=\\\".+\\\" logging_level=\\\"WARNING\\\" record=\\\"f\\\" ' \
+                r'required_option_1=\\\"value_1\\\" required_option_2=\\\"value_2\\\" show_configuration=\\\"f\\\"\"\]\]\}'
```
ljiang1, line 707: the only diff here is standardError vs. Exception? maybe a more clear if we do something like
inspect="....." (py2 version")
if !six.py2 (or just six.py3):
   inspect = inspect.replace("standardError", "exception")

tests/searchcommands/test_searchcommands_app.py
```diff
@@ -29,7 +29,22 @@
...
+try:
+    from itertools import imap  # python 2
+except ImportError:
+    from builtins import map as imap  # python 3
+
+try:
+    from itertools import izip  # python 2
+except ImportError:
+    from builtins import zip as izip  # python 3
+
```
ljiang1, line 47: instead of try-except, use if six.py2?

splunklib/searchcommands/eventing_command.py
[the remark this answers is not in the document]
ljiang1: if I read it correctly seems py3 only support items() and iteritems-> items in py3?

tests/searchcommands/test_internals_v1.py
amysutedja: Good call, will remove it.

PR #315, opened 2020-04-09
Mark tests as app

amysutedja:
Some tests require the [SDK App Collection](https://github.com/splunk/sdk-app-collection). If that set of apps does not exist, the test passes anyway but outputs a log message.

This changeset makes this requirement explicit. Failure to install the SDK App Collection will now cause test failures. You can exclude these tests by applying the new `app` Pytest marker.

This changeset also:

- Adds some helpful `make commands`
- Removes code that checks for Splunk versions below 5, which are no longer supported

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,20 @@ test:
 	@echo "$(ATTN_COLOR)==> test $(NO_COLOR)"
 	@tox -e py27,py37
 
+.PHONY: test_specific
+test_specific:
+	@echo "$(ATTN_COLOR)==> test_specific $(NO_COLOR)"
+	@sh ./scripts/test_specific.sh
+
 .PHONY: test_smoke
 test_smoke:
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -0,0 +1,55 @@
...
+# red
+ERROR_COLOR=\033[31;01m
+# cyan
+WARN_COLOR=\033[36;01m
+# yellow
+ATTN_COLOR=\033[33;01m
+
+ROOT_DIR := $(shell git rev-parse --show-toplevel)
+
+SPEC_REPO_BRANCH_OR_COMMIT := $(if $(SPEC_REPO_BRANCH_OR_COMMIT),$(SPEC_REPO_BRANCH_OR_COMMIT),latest)
```
fantavlik, line 14: I think this line should be removed, seems to be copypasted from another repo

PR #322, opened 2020-05-12
Travis/Tox CI Updates

amysutedja:
## Tox-Travis Integration

Uses `tox-travis` to correctly produce the build matrix.

## Splunk 7.3 container update

Uses the new Splunk 7.3 container, built from https://github.com/splunk/splunk-travis-ci-configs

At some point we should probably use https://github.com/Splunk/docker-Splunk and install the `sdk-app-collection` rather than constantly rebuilding containers.

## Tox changes for SearchCommand App Tests

It turns out that these tests are skipped entirely unless you've run a build command beforehand. Once you do that, it then turns out that the new Tox version of the run seems to have a problem writing to `$SPLUNK_HOME` which is set to be `/var/opt/Splunk/...`

Fix: Change the example apps to write to their own directory, where we know we always have write access.

## Python 3 changes for SearchCommand App Tests

Once you get past that hurdle, it then turns out that several tests fail in Python 3 because they're based on serialized recordings in Python 2, and Python 3 has different behavior. Some examples with respect to PY3 vs. PY2:

- More precision for some numerical values
- `\w` treats kanji characters as words

Fix: Recordings detect PY3-ness and try to look for a PY3 override file. If there isn't one, it falls back to the original PY2.

N.B.: One test is disabled for the moment while we figure out how to properly write that file.

## Makefile changes

Now `make` properly builds the SearchCommand app before running tests.


.travis.yml
```diff
@@ -29,19 +29,18 @@ before_install:
...
-  - "3.5"
+  - "3.7"
 
-install: "pip install unittest2"
+install: pip install tox-travis
 
 before_script: python setup.py build dist
 
-script: python setup.py test
+script: tox
```
shilpabijam, line 46: @amysutedja the README needs to be updated too corresponding to these changes?

tests/searchcommands/test_searchcommands_app.py
```diff
@@ -86,16 +86,20 @@ def __init__(self, path):
         splunk_cmd = path + '.splunk_cmd'
 
         try:
-            with io.open(splunk_cmd, 'rb') as f:
+            with io.open(splunk_cmd, 'r') as f:
```
shilpabijam, line 89: why the change from 'rb' to 'r'?

amysutedja: This is to fix a bug that manifests in Python 3. The line below attempts to call `.readline` on a binary-opened filestream, which conceptually you shouldn't be able to do. Python 2 allows you to do it because it has a very loose idea of what bytes and strings are, but it crashes in Python 3.

shilpabijam: Makes sense, thanks.

PR #324, opened 2020-05-22
Install sdk-app-collection via docker-compose

amysutedja:


docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3.6'
```
shakeelmohamed, line 1: can we move this file under the tests or scripts dir?

amysutedja: I'm thinking that we should keep it in the root because it seems like it's associated with `Dockerfile` and the recommendation is to keep that at root: https://stackoverflow.com/questions/27387811/where-to-keep-dockerfiles-in-a-project

Can you go a bit more into your reasoning here?

README.md
```diff
@@ -74,6 +75,17 @@ Alternatively, you can use **setup.py** on the sources you cloned from GitHub:
 However, it's not necessary to install the libraries to run the
 examples and unit tests from the SDK.
 
+## Testing Quickstart
+
+You'll need `docker` and `docker-compose` to get up and running using this method.
+
```
shilpabijam, line 81: Nitpick: can we have the `make up` in a new line.

README.md
```diff
@@ -74,6 +75,17 @@ Alternatively, you can use **setup.py** on the sources you cloned from GitHub:
...
+```
+SPLUNK_VERSION=8.0 make up
```
shilpabijam, line 83: Nitpick: can we have the `make up` in a new line.

docker-compose.yml
shakeelmohamed: This docker compose configuration stands up a Splunk Enterprise instance in Docker. It actually has nothing to do with the Splunk SDK for Python itself, it's used for testing.

I don't have a strong opinion, just wondering if that was a possibility

PR #331, opened 2020-06-16
Fix a couple of warnings spotted when running python 2.7 tests

xrmx:


splunklib/searchcommands/validators.py
```diff
@@ -95,10 +95,7 @@ def __call__(self, value):
         try:
             return Code.object(compile(value, 'string', self._mode), six.text_type(value))
         except (SyntaxError, TypeError) as error:
-            if six.PY2:
-                message = error.message
-            else:
-                message = str(error)
+            message = str(error)
```
ashah-splunk, line 98: @xrmx could you please help us with what was the warning message that was generated by this code?

xrmx: `DeprecationWarning: BaseException.message has been deprecated as of Python 2.6`

PR #337, opened 2020-08-14
Use module-specific logger in library code instead of root logger

majiang:
This pull request is to comply with [Advanced Logging Tutorial](https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial):

> A good convention to use when naming loggers is to use a module-level logger, in each module which uses logging, named as follows:
> ```python
> logger = logging.getLogger(__name__)
> ```

This pull request is also almost identical to #94 which I'm not sure why was closed.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -49,6 +49,7 @@ try:
 except ImportError as e:
     from xml.parsers.expat import ExpatError as ParseError
 
+logger = logging.getLogger(__name__)
 
 __all__ = [
     "AuthenticationError",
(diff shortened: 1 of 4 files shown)


delewis13:
+1 on this, using this module is adding a stream handler to my root level logger, which subsequently causes all my logging statements in the rest of my application to output multiple lines.

majiang:
Hi Splunk team, do you have any comments, thoughts or ideas about failing CI?

RCoff:
I definitely need this. The logs from 'binding.py' are unnecessary for me and only serve to clog up my Splunk index.

ashah-splunk:
@majiang as per your suggestion, we have added module specific logger along with an utility to enable loggers for different levels as well. Please refer the [README](https://github.com/splunk/splunk-sdk-python#optionalset-up-logging-for-splunklib)

majiang:
Thanks! It's #437, for reference

Issue #288, opened 2019-10-25
StreamingCommand failed when input contains non-ascii character


cui134:
Environment:
- splunk8.0
- splunk sdk 1.6.11
- python 3

When I create StreamingCommand   "testcommand" with following code:

```
#!/usr/bin/env python
#-*- coding: utf-8 -*-
import sys
from splunklib.searchcommands import \
    dispatch, StreamingCommand, Configuration, Option, validators
import splunk
@Configuration()   
class TestCommand(StreamingCommand):
    def stream(self, events):   
        for event in events:        
            yield event
dispatch(TestCommand, sys.argv, sys.stdin, sys.stdout, __name__)
```

testcommand will hang when I use following SPL:
sourcetype=XXX| search url = "http://例子.卷筒纸" | testcommand

There has follwing error log in splunkd.log:

> 10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: During handling of the above exception, another exception occurred:
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: Traceback (most recent call last):
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: File "C:\Program Files\Splunk\etc\apps\XXXX\bin\testcommand.py", line 22, in
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: dispatch(TestCommand, sys.argv, sys.stdin, sys.stdout, name)
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: File "C:\Program Files\Splunk\etc\apps\XXXX\bin\splunklib\searchcommands\search_command.py", line 1118, in dispatch
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: command_class().process(argv, input_file, output_file)
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: File "C:\Program Files\Splunk\etc\apps\XXXX\bin\splunklib\searchcommands\search_command.py", line 435, in process
10-25-2019 16:34:42.808 +0800 ERROR ChunkedExternProcessor - stderr: self._process_protocol_v2(argv, ifile, ofile)
[...]

PR #341, opened 2020-09-04
Custom search command support for multibyte characters in Python 3

amysutedja:
Fixes https://github.com/splunk/splunk-sdk-python/issues/288

## `SearchCommand` supports multibyte characters in Python 3

Previously, `SearchCommand` in Python 3 would read directly from the incoming `ifile` stream -- typically `sys.stdin`. In Python 2 `sys.stdin` is a file-like byte stream, whereas in Python 3 it is an `io.TextIOWrapper` containing an underlying buffer. Because this object reads by character rather than by byte, multibyte characters would cause the command to read too far past the data's boundary. This could lead to corrupt data reads (if early in the stream) or infinite hangs (if at the end of the stream).

We now retrieve the underlying buffer and read from it when in Python 3. The read bytes are then cast to strings for parsing purposes.

## Tests ensure underlying byte stream

Previously, the tests defined a metadata stream with the `Ａ` character in it (not to be confused with `A`). In Python 2, this character caused its containing string to become `unicode`, which caused `StringIO` to gain that encoding. As a result, the size of the metadata stream was always incorrectly measuring Unicode characters rather than bytes, but under test the read logic would always be handed a Unicode character stream rather than a byte stream.

This has been fixed.

## Multibyte test fixture

We now have a new test `test_multibyte_chunked` which contains a multibyte character test fixture.


ichaer-splunk:
Hah, I've been working since yesterday in a change that is nearly identical to this. Nice fix, if I say so myself =D

I'll retire my fork, this is really good.

splunklib/searchcommands/search_command.py
```diff
@@ -852,15 +852,23 @@ def _execute(self, ifile, process):
     @staticmethod
     def _read_chunk(ifile):
         # noinspection PyBroadException
+        if six.PY2:
+            istream = ifile
+        else:
+            try:
+                istream = ifile.buffer
```
aij, line 859: I think this logic could be moved to `_process_protocol_v2` before calling `_read_chunk`, since it doesn't need to be re-evaluated for each chunk.

Also, since intermixing reads from `sys.stdin` and `sys.stdin.buffer` does not work well (presumably due to buffering, but I haven't checked), it might be safer to use`ifile = ifile.buffer` so that the wrong variable can't be used by mistake. (It also means the assignment for python2 can be skipped entirely.)

tests/searchcommands/test_search_command.py
```diff
@@ -33,8 +33,21 @@
 import os
 import re
 
+from io import TextIOWrapper
+
 import pytest
 
+def build_command_input(getinfo_metadata, execute_metadata, execute_body):
+    header = ('chunked 1.0,{},0\n{}'.format(len(six.ensure_binary(getinfo_metadata)), getinfo_metadata) +
```
aij, line 41: `header` seems like an odd name for the string that contains the entire command input. 

I really like seeing this split out into a function though. :+1:

aij (approved):
Thanks for fixing this so quickly!

I added a couple comments, but nothing worth blocking on.

aij (approved):

tests/searchcommands/test_search_command.py
amysutedja: Renamed.

splunklib/searchcommands/search_command.py
amysutedja: Moved this logic into `_as_binary_stream` and created a Python 3-only assert to ensure that incoming objects come from the binary stream hierarchy.

zenmoto (approved):

tristan-splunk:
FYI we are waiting on this change to upgrade the pythonsdk in all our apps ahead of conf. We'd rather not have to do a monkey patch. Please advise when you can merge and release an official build.

PR #342, opened 2020-09-04
wip Preserve errors so they will be logged to stderr with backtraces

aij:
Hiding exceptions just makes debugging a pain.

Unfortunately, there are a number of tests which try to run the
search command within the same process, leading them to assert
current implementation details like SystemExit exceptions.


Ickerday:

PR #350, opened 2020-10-06
initial end-to-end tests for streaming, reporting, generating

zenmoto:
Start on end-to-end tests for search commands

### DIFF
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -851,7 +851,8 @@ class SearchCommand(object):
 
     @staticmethod
     def _as_binary_stream(ifile):
-        if six.PY2:
+        naught = ifile.read(0)
+        if isinstance(naught, bytes):
             return ifile
 
         try:
--- a/tests/searchcommands/chunked_data_stream.py
+++ b/tests/searchcommands/chunked_data_stream.py
@@ -42,11 +42,11 @@ class ChunkedDataStream(collections.Iterable):
     def read_chunk(self):
         header = self.stream.readline()
 
-        while header > 0 and header.strip() == b'':
+        while len(header) > 0 and header.strip() == b'':
             header = self.stream.readline()  # Skip empty lines
-
-        if not header == 0:
+        if len(header) == 0:
             raise EOFError
+
         version, meta, data = header.rstrip().split(b',')
         metabytes = self.stream.read(int(meta))
         databytes = self.stream.read(int(data))
--- a/tests/searchcommands/chunked_data_stream.py
+++ b/tests/searchcommands/chunked_data_stream.py
@@ -39,11 +39,12 @@ class ChunkedDataStream(collections.Iterable):
 
     def read_chunk(self):
         header = self.stream.readline()
-        if len(header) == 0:
-            raise EOFError
 
         while len(header) > 0 and header.strip() == b'':
             header = self.stream.readline()  # Skip empty lines
+
+        if len(header) == 0:
+            raise EOFError
         version, meta, data = header.rstrip().split(b',')
         metabytes = self.stream.read(int(meta))
         databytes = self.stream.read(int(data))
@@ -82,22 +83,13 @@ def _build_data_csv(data):
         return b''
     if isinstance(data, bytes):
         return data
-    if six.PY2:
-        csvout = io.BytesIO()
-    else:
-        csvout = io.StringIO()
+    csvout = splunklib.six.StringIO()
 
     headers = set()
     for datum in data:
-        if six.PY2:
-            headers.update(datum.keys())
-        else:
-            headers.update(datum.keys())
+        headers.update(datum.keys())
     writer = csv.DictWriter(csvout, headers, dialect=splunklib.searchcommands.internals.CsvDialect)
     writer.writeheader()
     for datum in data:
-        if six.PY2:
-            writer.writerow(datum)
-        else:
-            writer.writerow(datum)
+        writer.writerow(datum)
     return six.ensure_binary(csvout.getvalue())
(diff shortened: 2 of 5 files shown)


tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+        header = self.stream.readline()
+        if len(header) == 0:
+            raise EOFError
+
+        while len(header) > 0 and header.strip() == b'':
+            header = self.stream.readline()  # Skip empty lines
+        version, meta, data = header.rstrip().split(b',')
+        metabytes = self.stream.read(int(meta))
+        databytes = self.stream.read(int(data))
+        print("sent")
```
amysutedja, line 50: Should probably be removed or go to `stderr`.

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+        return ChunkedDataStreamIter(self)
+
+    def __init__(self, stream):
+        empty = stream.read(0)
+        assert isinstance(empty, bytes)
+        self.stream = stream
+
+    def read_chunk(self):
+        header = self.stream.readline()
+        if len(header) == 0:
```
amysutedja, line 42: Should this be moved after the "skip empty lines" logic?

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+        version, meta, data = header.rstrip().split(b',')
+        metabytes = self.stream.read(int(meta))
+        databytes = self.stream.read(int(data))
+        print("sent")
+        return Chunk(version, metabytes, databytes)
+
+
+def build_chunk(kv, data=None):
+    metadata = six.ensure_binary(json.dumps(kv), 'utf-8')
+    data_output = _build_data_csv(data)
```
amysutedja, line 56: IMO we should also have `six.ensure_binary` here as well.

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+    return b"chunked 1.0,%d,%d\n%s%s" % (len(metadata), len(data_output), metadata, data_output)
+
+
+def build_empty_searchinfo():
+    return {
```
amysutedja, line 61: Not a constant?

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+    return build_chunk({'action': 'execute', 'finished': finished}, data)
+
+
+def _build_data_csv(data):
+    if data is None:
+        return b''
+    if isinstance(data, bytes):
+        return data
+    if six.PY2:
+        csvout = io.BytesIO()
```
amysutedja, line 86: I think `six` has an alias that makes `BytesIO` do the right thing here.

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        csvout = io.StringIO()
+
+    headers = set()
+    for datum in data:
+        if six.PY2:
```
amysutedja, line 92: Feels like we can remove this `if` statement and the one below.

tests/searchcommands/chunked_data_stream.py
```diff
@@ -0,0 +1,103 @@
...
+            headers.update(datum.keys())
+        else:
+            headers.update(datum.keys())
+    writer = csv.DictWriter(csvout, headers, dialect=splunklib.searchcommands.internals.CsvDialect)
+    writer.writeheader()
+    for datum in data:
+        if six.PY2:
+            writer.writerow(datum)
```
amysutedja, line 100: I'm probably missing something, but how do we know that the data values align with the order of the headers if we're making a set of potentially different-looking data?

amysutedja (changes requested):

splunklib/searchcommands/search_command.py
```diff
@@ -851,7 +851,8 @@ def _execute(self, ifile, process):
 
     @staticmethod
     def _as_binary_stream(ifile):
-        if six.PY2:
+        naught = ifile.read(0)
+        if isinstance(naught, bytes):
```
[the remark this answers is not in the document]
zenmoto: Well, got it from you! :-)

tests/searchcommands/chunked_data_stream.py
zenmoto: D'oh.

tests/searchcommands/chunked_data_stream.py
zenmoto: In this case _build_data_csv explicitly returns a binary- in fact the last line of that function is `return six.ensure_binary(csvout.getvalue())`

tests/searchcommands/chunked_data_stream.py
zenmoto: The idea here was to be able to parameterize this as needed to build custom searchinfo objects, but I didn't need anything more than these fields existing yet.

tests/searchcommands/chunked_data_stream.py
zenmoto: Oh, yeah- I did get those unified. Thanks, I'll remove that.

tests/searchcommands/chunked_data_stream.py
zenmoto: This is being managed by the `csv.DictWriter`. This looks like another place I had to take different paths for the pythons but then were able to re-converge them.

tests/searchcommands/chunked_data_stream.py
zenmoto: Took me a second, but yes, you are right.

amysutedja (approved):
Approved pending passing tests.

PR #352, opened 2020-10-16
Allow suppying of a key-value body when calling Context.post()

zenmoto:

PR #346, opened 2020-09-17
Readme updates

apruneda:
Changes:
- Did an overall edit of the README and CONTRIBUTING pages
- Updated URLs
- Updated branding (Splunk Enterprise SDK)
- Fixed formatting issues

PR #364, opened 2021-01-28
Release v1.6.15

fantavlik:
## Bug fixes
#301 Fix chunk synchronization
#327 Rename and cleanup follow-up for chunk synchronization
#352 Allow supplying of a key-value body when calling Context.post()

## Minor changes
#350 Initial end-to-end tests for streaming, reporting, generating custom search commands
#348 Update copyright years to 2020 
#346 Readme updates to urls, terminology, and formatting
#317 Fix deprecation warnings


splunklib/searchcommands/internals.py
```diff
@@ -748,37 +772,36 @@ class RecordWriterV2(RecordWriter):
...
 
-        self._finished = finished is True
+    def write_chunk(self, finished=None):
+        inspector = self._inspector
+        self._committed_record_count += self.pending_record_count
+        self._chunk_count += 1
+
+        # TODO: DVPL-6448: splunklib.searchcommands | Add support for partial: true when it is implemented in
+        # ChunkedExternProcessor (See SPL-103525)
+        #
```
ljiang1, line 791: do we need to remove the bug#?

ljiang1 (approved):

splunklib/searchcommands/internals.py
fantavlik: Do you mean you think the bug has addressed and we can remove this comment? We are OK to refer to ticket numbers as long as we don't provide links if that's the concern.

fantavlik:
FYI there was one failing test which is known to be flaky for 8.x Splunk versions.

PR #371, opened 2021-03-13
Modinput preserve 'app' context

lowell80:
Add the ability to track the `app` namespace for modular inputs.  This information is available in the XML passed in via stdin, but has not been available to modular input code using splunklib.  Having access to app allows additional interactions with the REST API to be properly namespaces, for example if you need to lookup a password or other conf entry that lives in the same namespace as the input stanza.

Example use case with the updated code:

```python
def stream_events(self, inputs, ew):
    for input_name, input_item in inputs.inputs.items():
        app = self.service.namespace["app"] = input_item["__app"]
        my_pass = self.service.storage_passwords["my_password"]
```

I picked the name `__app` to avoid potential naming conflicts with entries in `inputs.conf`, but it's a bit arbitrary.  I'd be happy to change it if there are strong preferences to the contrary.


bparmar-splunk:
Hi @lowell80,
We have tried to implement these changes but we have encountered some errors while executing test cases.
We look forward to review it again, if you push these changes (including conflicts resolved) with regards to newer sdk version.

Also, test cases will be automatically triggered (using github workflow) when changes are pushed. So it would help us to resolve, if any error occurs.

lowell80:
@bparmar-splunk, I've rebased my branch on the develop branch, but some of the test are failing.  I suspect this is related to things outside the changes I made.  Can you take a look at the test logs and determine if this is an issue with my code?

Between action job 182 (incorrect rebase mistake by me) and 183 (rebase on develop) I see the errors:
* `[Errno 104] Connection reset by peer`
* `HTTP 500 Internal Server Error -- already reloading or shutting down, will not reload`

bparmar-splunk:
Hi @lowell80,
I have re-initiated test run and all checks are passed. Hence, all code looks good.

PR #372, opened 2021-03-30
Update __init__.py

hogoboom:
Add support for Splunk authentication tokens in example scripts (search.py, etc.)


ashah-splunk:
@hogoboom , in the latest release we have added support for both the authentication tokens i.e splunkToken(sessionKey) and token(bearer token). please refer [updated](https://github.com/splunk/splunk-sdk-python/blob/master/utils/__init__.py#L73). Closing this PR as the change is already been added.

PR #377, opened 2021-05-21
Fix  #373: add bool() defninition

### LINKED ISSUE #373: search.py verbose flag issue
I do understand there are still ongoing works to migrate python2 to python3
Just wanted to document `example/search.py`

```bash
Traceback (most recent call last):
  File "splunk-sdk-python\examples\search.py", line 118, in <module>
    main(sys.argv[1:])
  File "splunk-sdk-python\examples\search.py", line 96, in main
    if verbose > 0:
TypeError: '>' not supported between instances of 'str' and 'int'
```

This can be interpreted to:
```python
if bool(verbose) > 0:
    ...
```
 [bool()](https://docs.python.org/3.8/library/functions.html?highlight=bool#bool) defninition

deloboiden:


ashah-splunk:
Hi @deloboiden , thanks for the PR, but we tried executing the example mentioned in the Issue #373 and it worked fine in both the scenarios with and without the verbose flag for both python2 and python3, so the suggested change is not required. Hence closing the PR

Issue #380, opened 2021-06-10
Parameter for additional verifiy locations for Server Certificates during connect


AndyXan:
**Is your feature request related to a problem? Please describe.**
The Splunk Python SDK lacks the possibility to provide a custom ca certificate that is not installed in the local system certificates storage for server certificate validation. 

The possible arguments (key_file, cert_file) are for **client authentifaction** - not for server certificate verficitation. 

The problematic Code snippet is (current upstream): 
binding.py::1354

```python
def handler(key_file=None, cert_file=None, timeout=None, verify=False):
    """This class returns an instance of the default HTTP request handler using
    the values you provide.

    :param `key_file`: A path to a PEM (Privacy Enhanced Mail) formatted file containing your private key (optional).
    :type key_file: ``string``
    :param `cert_file`: A path to a PEM (Privacy Enhanced Mail) formatted file containing a certificate chain file (optional).
    :type cert_file: ``string``
    :param `timeout`: The request time-out period, in seconds (optional).
    :type timeout: ``integer`` or "None"
    :param `verify`: Set to False to disable SSL verification on https connections.
    :type verify: ``Boolean``
    """

    def connect(scheme, host, port):
        **kwargs = {} <-----**
        if timeout is not None: kwargs['timeout'] = timeout
        if scheme == "http":
            return six.moves.http_client.HTTPConnection(host, port, **kwargs)
        if scheme == "https":
            if key_file is not None: kwargs['key_file'] = key_file
            if cert_file is not None: kwargs['cert_file'] = cert_file

            if not verify:
                kwargs['context'] = ssl._create_unverified_context()
            
            return six.moves.http_client.HTTPSConnection(host, port, **kwargs) <-----------------
        raise ValueError("unsupported scheme: %s" % scheme)
```

So, if verification is enabled - a default SSL Context is created and CA's `.pem` file have to be present in the system certificate storage.  Otherwise it's an error like that: 

```python
import http.client
c = http.client.HTTPSConnection("my-splunk-host", "8000")
c.request("GET", "/")

[Out] SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] 
```

What's actually desired is to provide additional ssl lookup definitions: 

```python
import ssl
import http.client
context = ssl.create_default_context()
context.load_verify_locations("my-root-ca.pem")
c = http.client.HTTPSConnection("my-splunk-host", "8000", context=context)
test = c.request("GET", "/")
```

[...]

PR #383, opened 2021-06-14
#380 Implemented the possibility to provide a SSLContext object to the connect method

AndyXan:
#380 Implemented the possibility to provide a SSLContext object to the connect method


fantavlik (approved):
Thanks for the contribution @AndyXan - this looks good to us.

PR #392, opened 2021-08-29
Break out search argument to option parsing for v2 custom search commands

acharlieh:
So I was wanting to write a custom search command, but for reasons of making things similar to existing commands, and because what I'm needing to take as positional parameters could have equals signs in them, I wanted to parse options that look like `--optionName=optionValue` instead of `optionName=optionValue`

Digging through the SDK I think I found the point where we were doing this, and wrapped it in a method so that I could override this parsing, but keep everything else passive for others who are already using the SDK.

I could see a future extension point where this entire parsing out options might be able to be overridden/modified, but I didn't have a need for that much customization yet.

### DIFF
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -634,6 +634,19 @@ class SearchCommand(object):
 
         debug('%s.process finished under protocol_version=1', class_name)
 
+    def _protocol_v2_option_parser(self, arg):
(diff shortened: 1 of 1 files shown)


acharlieh:
Helps if I make the pull request against the correct branch.

fantavlik:
Looks good to me since there are no functional changes for end users, good luck @acharlieh

PR #397, opened 2021-09-22
Fix for Issue#361

rmaheshwari-splunk:
Added code changes for encoding '/' in _key parameter in kvstore.data APIs.


fantavlik (approved):
Looks good, thanks for investigating.

Issue #356, opened 2020-11-02
KV Store Collection API - dict Type Issue 


malvidin:
**Describe the bug**
The `**query` in KV Store Data API functions have an error in typing or in usage in the function.
https://github.com/splunk/splunk-sdk-python/blob/9f6b2e041e7e4d09c70cc32c02d25c694842bce7/splunklib/client.py#L3633
https://github.com/splunk/splunk-sdk-python/blob/9f6b2e041e7e4d09c70cc32c02d25c694842bce7/splunklib/client.py#L3643

**To Reproduce**
Use the `KVStoreCollectionData` query with the dict type prescribed in the docstrings
`collection.data.query(query={"id": {"$gt": 24}}) or collection.data.query({"id": {"$gt": 24}})`

**Expected behavior**
The previous queries should succeed, as documented.
https://docs.splunk.com/Documentation/Splunk/8.1.0/RESTREF/RESTkvstore

***Splunk Version***
- Version: 8.1.0
- OS: Docker Build 5c3f219bbb6b

***splunklib Version***
 - Version: 1.6.14
 - Language Runtime Version: Python 3.7
 - OS: Ubuntu 20.04

akaila-splunk:
Hi @malvidin , Thanks for bringing it up.
We have considered your suggestion and added dictionary support for KVStore "query" methods. Also updated the [KVStore example](https://github.com/splunk/splunk-sdk-python/blob/develop/examples/kvstore.py) file for reference. Changes will be available in the next release.

ashah-splunk:
Hi @malvidin , we have a new SDK release 1.6.17 with the proposed changes. Would request you to try with the latest SDK version. Let us know in case if you face any issue with the change implemented.

PR #398, opened 2021-09-22
Updated kvstore method

akaila-splunk:
Added dictionary support for KVStore "query" methods. Change with reference to issue #356 . KVStore example also updated


fantavlik (approved):
Looks good, thanks for updating the examples too that's perfect

PR #399, opened 2021-09-24
Move CI tests in Python SDK to Github Actions from TravisCI

akaila-splunk:
- Added test.yml file for test cases execution on push and PR
- Remove .travis.yml file


fantavlik (changes requested):
In general these changes look good and I don't expect all tests to pass at this point but a number of the jobs are terminating with: ```Error: The operation was canceled.``` could you possibly investigate those issues before merging this work? We'll also want to followup once we know which tests are failing and fix or skip those tests after investigating but that can be a followup task. For now I just want to make sure that all tests are being executed and not running into the above error.

akaila-splunk:
GitHub cancels all in-progress jobs with error `Error: The operation was canceled` if any matrix job fails. that's the reason we are getting this error for other jobs when one of the job fails. 
I have changed code as request and now all the jobs are running even if any job fails.

fantavlik:
> GitHub cancels all in-progress jobs with error `Error: The operation was canceled` if any matrix job fails. that's the reason we are getting this error for other jobs when one of the job fails.
> I have changed code as suggested and now all the jobs are running even if any job fails.

Excellent, that's good to know - thanks for investigating.

fantavlik (approved):
Awesome work - let's merge this. Can you also create a followup ticket to address the test failures now that we can see those? That will be high priority.

akaila-splunk:
> Awesome work - let's merge this. Can you also create a followup ticket to address the test failures now that we can see those? That will be high priority.

Sure, I'll create ticket for test failures.

PR #407, opened 2021-10-27
Modified Streaming and Generating Custom Search Command

akaila-splunk:
- Fix field dropping issue for Streaming and Generating CSC

### DIFF
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -179,7 +179,7 @@ class SearchCommand(object):
 
     def gen_record(self, **record):
         self._record_writer.custom_fields |= record.keys()
-        return {**record}
+        return record
 
     record = Option(doc='''
         **Syntax: record=<bool>
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -178,7 +178,7 @@ class SearchCommand(object):
         current_record[field_name] = field_value
 
     def gen_record(self, **record):
-        self._record_writer.custom_fields |= record.keys()
+        self._record_writer.custom_fields |= set(record.keys())
         return record
 
     record = Option(doc='''
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -177,6 +177,10 @@ class SearchCommand(object):
         self._record_writer.custom_fields.add(field_name)
         current_record[field_name] = field_value
 
+    def gen_record(self, **record):
+        self._record_writer.custom_fields |= record.keys()
+        return {**record}
+
     record = Option(doc='''
         **Syntax: record=<bool>
 
--- a/splunklib/searchcommands/search_command.py
+++ b/splunklib/searchcommands/search_command.py
@@ -176,7 +176,6 @@ class SearchCommand(object):
     def add_field(self, current_record, field_name, field_value):
         self._record_writer.custom_fields.add(field_name)
         current_record[field_name] = field_value
-        return current_record
 
     record = Option(doc='''
         **Syntax: record=<bool>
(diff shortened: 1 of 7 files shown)


tests/searchcommands/test_generator_command.py
```diff
@@ -59,3 +58,29 @@ def generate(self):
     except ValueError as error:
         assert str(error) == "allow_empty_input cannot be False for Generating Commands"
 
+def test_all_fieldnames_present_for_generated_records():
```
fantavlik, line 61: This is great, can we add a similar test for `add_field` for streaming commands?

fantavlik (changes requested):
This approach is really looking good, thanks for the documented examples and generating command test! Just one request to add a full fledged tests for the `add_field` scenario to ensure that is working as intended.

fantavlik (changes requested):
This approach is really looking good, thanks for the documented examples and generating command test! Just one request to add a full fledged tests for the `add_field` scenario to ensure that is working as intended.

splunklib/searchcommands/search_command.py
```diff
@@ -173,6 +173,14 @@ def logging_level(self, value):
                 raise ValueError('Unrecognized logging level: {}'.format(value))
         self._logger.setLevel(level)
 
+    def add_field(self, current_record, field_name, field_value):
+        self._record_writer.custom_fields.add(field_name)
+        current_record[field_name] = field_value
```
Bre77, line 178: This function should return current_record or else the documentation is incorrect.
```suggestion
    def add_field(self, current_record, field_name, field_value):
        self._record_writer.custom_fields.add(field_name)
        current_record[field_name] = field_value
        return current_record
```

fantavlik: Good catch @Bre77 - I think in this case the documentation needs to be updated, the `current_record` should be updated with the new field without the need to return since it's essentially pass-by-reference for this usage (I always go back to this article when I need to refresh my memory: https://robertheaton.com/2014/02/09/pythons-pass-by-object-reference-as-explained-by-philip-k-dick/).

So if you all agree can we fix the README.md above to the following?
```
class CustomStreamingCommand(StreamingCommand):
    def stream(self, records):
        for index, record in enumerate(records):
            if index % 1 == 0:
                self.add_field(record, "odd_record", "true")
            yield record
```

fantavlik: And to make sure that example we provide is working ^ that's probably a good implementation to add when adding tests for the StreamingCommand / add_field case

Bre77: I've changed my implementation to using pass by reference (as per the suggested doco update) and everything works as expected.

vmalaviya-splunk: Good catch @Bre77 
I've updated the README accordingly.

tests/searchcommands/test_generator_command.py
vmalaviya-splunk: Absolutely.
I have added a test for the streaming command.

tests/searchcommands/test_streaming_command.py
```diff
@@ -27,3 +27,71 @@ def stream(self, records):
...
+    ifile.seek(0)
+    ofile = io.BytesIO()
+    cmd._process_protocol_v2([], ifile, ofile)
+    ofile.seek(0)
+    output_iter = chunky.ChunkedDataStream(ofile).__iter__()
+    output_iter.next()
+    output_records = [i for i in output_iter.next().data]
+
+    # Assert that count of records having "odd_field" is 0
+    assert len(list(filter(lambda r: "odd_field" in r, output_records))) == 0
```
fantavlik, line 60: Really nice demonstration of the issue!

fantavlik (approved):
Looking forward to getting this out to our eager users 🚀 nice work

PR #413, opened 2021-11-08
Default kvstore owner to nobody

ashah-splunk:
Added a new property for the kvstore owner, which is set to "nobody" by default. But a setter is added for future purpose when the owner value is required to be changed


fantavlik (approved):
Looks good, thanks for updating testcases for both the default case (nobody) and override case (admin) 🚀

Issue #400, opened 2021-09-24
`service.storage_passwords.list` misses space in `username` field


arys-splunk:
**Describe the bug**
`service.storage_passwords.list` misses space in `username` field.

**To Reproduce**
1. Install https://splunkbase.splunk.com/app/2731/
2. Add 
```
[ test]
account_name = ucspe
account_password = ucspe
disable_ssl_verification = True
server_url = 10.141.36.107
```
to `local/splunk_ta_cisco_ucs_servers.conf`.
3. Go to Splunk, open Cisco UCS app and refresh the page.
4. Run this Python script
```python
from splunklib.client import connect

service = connect(username="admin", password="<your-admin-password>")


all_passwords = service.storage_passwords.list(count=-1, search="")

for p in all_passwords:
    print(p.state)
```
5. See that username field does not have space, but the stanza that we add in step 2 has.

**Expected behavior**
Username field should have space in it.

**Splunk (please complete the following information):**
- Version: 8.2.2.1
- OS: MacOS 11.6
- Deployment: single-instance

**SDK (please complete the following information):**
 - Version: 1.6.16
 - Language Runtime Version: Python 3.7
 - OS: MacOS 11.6

akaila-splunk:
Hi @arys-splunk , We have fixed the issue, and the changes will be available in the next release.
PR for reference - #416

ashah-splunk:
@arys-splunk please refer the latest python-sdk release 1.6.19, and please let us know if the issue still exists.

PR #416, opened 2021-11-19
Removed strip() method in load_value() method from data.py file

akaila-splunk:
- removed strip() method in load_value() method, so now response value contain leading and trailing spaces while printing data in console. Change with reference to issue #400

### DIFF
--- a/splunklib/data.py
+++ b/splunklib/data.py
@@ -161,8 +161,8 @@ def load_value(element, nametable=None):
         text = element.text
         if text is None: 
             return None
-        text = text.strip()
-        if len(text) == 0: 
+
+        if len(text.strip()) == 0:
             return None
         return text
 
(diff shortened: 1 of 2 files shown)


arys-splunk (commented):
What do you think about adding test for this fix?

akaila-splunk:
> What do you think about adding test for this fix?

Hi @arys-splunk , Added new test case for the fix, let us know if any other changes are required

arys-splunk:
> > What do you think about adding test for this fix?
> 
> Hi @arys-splunk , Added new test case for the fix, let us know if any other changes are required

I am not familiar with the code, let's wait what @zenmoto and @fantavlik say.

fantavlik (approved):
Looks good, nice catch @akaila-splunk

Issue #345, opened 2020-09-13
Endpoint instance to HEC fails with 404 not found


dtrodger:
Calls to this endpoint return a 404. The SDK appends a '/' to the end of any uri. So, /services/collector/event is found, /services/collector/event/ is not.

`event_collector_endpoint = splunklib.client.Endpoint(splunk_service, "/services/collector/event")`

Subclassing client.Endpoints and overriding init like below works.

<img width="622" alt="Screen Shot 2020-09-13 at 7 58 03 PM" src="https://user-images.githubusercontent.com/16841791/93033329-7367cb80-f5fb-11ea-9aa3-fd4ba697614e.png">

ashah-splunk:
Hi @dtrodger , We have fixed the issue, and the changes will be available in the next release.

PR #419, opened 2021-11-26
Hec endpoint issue

ashah-splunk:
Fix for the Issue #345 

Previously `path` was always updated to `path + /`
This implementation fails while using the endpoint `/services/collector/event` as `/services/collector/event/` results in 404

The fix removes the hard coded addition of the `/` to all the paths, and it makes it conditional based on the presence of the `path_segment` for the `get/post` actions. If `path_segment' exists the `path` will be updated to `path + / `

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -724,7 +724,7 @@ class Endpoint(object):
     """
     def __init__(self, service, path):
         self.service = service
-        self.path = path #if path.endswith('/') else path + '/'
+        self.path = path
 
     def get(self, path_segment="", owner=None, app=None, sharing=None, **query):
         """Performs a GET operation on the path segment relative to this endpoint.
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -724,7 +724,7 @@ class Endpoint(object):
     """
     def __init__(self, service, path):
         self.service = service
-        self.path = path if path.endswith('/') else path + '/'
+        self.path = path #if path.endswith('/') else path + '/'
 
     def get(self, path_segment="", owner=None, app=None, sharing=None, **query):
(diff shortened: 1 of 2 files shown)


splunklib/client.py
```diff
@@ -842,6 +844,8 @@ def post(self, path_segment="", owner=None, app=None, sharing=None, **query):
         if path_segment.startswith('/'):
             path = path_segment
         else:
+            if not self.path.endswith('/') and path_segment != "":
+                self.path = self.path if path_segment.startswith('/') else self.path + '/'
```
fantavlik, line 848: Can you explain the fix? I'm not quite sure just by looking at this line why the path starting with `/` here `if path_segment.startswith('/')` would have an affect on whether an ending `/` would be needed and this line seems to be adding a `/` slash whereas line 727 was seemingly doing the same before? In general it's nice to get some sort of writeup (a few sentences) in the PR summary as to what was broken before and how the behavior changed. For this case you could give an example of how the old path was formed and what the new HEC path is now and how other paths are affected after this change.

fantavlik (changes requested):
Just need a little more context for what the change is doing and how other paths are affected.

splunklib/client.py
ashah-splunk: Previously `/` was added always for all the paths irrespective of the path_segment, hence the resulting path was always `path = path + /` which failed for the HEC as `/services/collector/event/` results in 404, 
Now `/` will be added only when `path_segment` is present so that the resulting `path = path + / + path_segment` and if there is no path_segment then the resulting `path = path` 
Now the `/` is added conditionally and hence removed from line 727
I will add some comments and also add a test case for the HEC and update the PR

fantavlik (approved):

PR #429, opened 2022-01-18
documented how to access modular input metadata

akaila-splunk:
- modified README.md file and github_commit.py file so user can easily get an idea of how to access modular input metadata
- gave reference of github_commits modular input example for better understanding


fantavlik (approved):
🎉

PR #433, opened 2022-01-28
Use "Apache 2.0" instead of a link in setup.py

### LINKED ISSUE #432: Change license to Apache 2.0 instead of using a link
`setup.py` file https://github.com/splunk/splunk-sdk-python/blob/56c108b0ccbbbb9470b4d3ba721ef41204214008/setup.py#L140 describes license as "http://www.apache.org/licenses/LICENSE-2.0", which shows as a link on PyPI website (see screenshot attached)
image

artemrys:
Closes #432

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -137,7 +137,7 @@ setup(
 
     description="The Splunk Software Development Kit for Python.",
 
-    license="http://www.apache.org/licenses/LICENSE-2.0",
+    license="Apache 2.0",
 
     name="splunk-sdk",
 


vmalaviya-splunk:
Thanks, @artemrys, for the PR. May I know which issue this PR addresses?

artemrys:
> Thanks, @artemrys, for the PR. I would like to know more about the issue this PR addresses.

This is a style type of change, I searched through some packages on PyPI and mostly people don't use links in the license field.

But if you think it is not needed and you want just to close it, feel free to do it :)

PR #437, opened 2022-03-03
added setup_logging() method in splunklib for logging and updated README.md files

akaila-splunk:
- Used module-specific logger in splunklib code instead of root logger
- Added new setup_logging() method in splunklib to set logging level, log format(optional) and date format(optional)
- Updated REAME.md file with details of setup_logging() method
- Updated searchcommands_app/README.md file with some details of logging.conf file and how to see debug logs
- Here is an example of log message,
```
Before code change:
	DEBUG:root:Log message
After code change:
	2022-03-03 19:12:58 India Standard Time, Level=DEBUG, Pid=11120, Logger=splunklib.binding, File=binding.py, Line=72, Log message
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,6 +243,19 @@ class GeneratorTest(GeneratingCommand):
         checkpoint_dir = inputs.metadata["checkpoint_dir"]
 ```
 
+#### Optional:Set up logging for splunklib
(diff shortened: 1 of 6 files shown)


fantavlik (approved):
Well done! Thanks for walking us through the design along the way on our weekly syncs

PR #439, opened 2022-03-08
Modified POST method debug log to not log sensitive body/data

akaila-splunk:
- This PR is fixed for the issue - 'splunklib logs user credentials in raw text'
- Added check to avoid writing sensitive data in debug logs
- ex. '/storage/passwords' endpoint is having password field in it's body during post method call


fantavlik (approved):
🎉

PR #441, opened 2022-03-15
JSONResultsReader added and deprecated ResultsReader 

ashah-splunk:
- Added user's approach for JSONResultsReader in place of ResultsReader 
- Newer approach increase efficiency and performance 
- Modified test cases, examples and documentation to use JSONResultsReader along with query param 'output_mode' set to 'json' to use the newer approach
- Added deprecation warning for the ResultsReader

### DIFF
--- a/examples/follow.py
+++ b/examples/follow.py
@@ -42,7 +42,7 @@ def follow(job, count, items):
             job.refresh()
             continue
         stream = items(offset+1)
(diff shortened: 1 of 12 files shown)


fantavlik (approved):
This looks good! One minor comment - the `ResultsReader` import in `examples/stail.py` can be removed.

PR #442, opened 2022-03-21
Optional retries feature added

ashah-splunk:
- Added optional retries feature provided by user in [PR](https://github.com/splunk/splunk-sdk-python/pull/293)
- Added test case for the optional retries
- Added example for the optional retries
- Added optional retry parameter for existing test cases setup, to prevent CI build failures


splunklib/binding.py
```diff
@@ -453,6 +455,11 @@ class Context(object):
     :type splunkToken: ``string``
     :param headers: List of extra HTTP headers to send (optional).
     :type headers: ``list`` of 2-tuples.
+    :param retires: Number of retries for each HTTP connection (optional, the default is 0).
+                    NOTE THAT THIS MAY INCREASE THE NUMBER OF ROUND TRIP CONNECTIONS TO THE SPLUNK SERVER.
```
fantavlik, line 459: One minor note, it's probably worth noting that this is also done in a blocking/synchronous way, so maybe amend this to:
```
NOTE THAT THIS MAY INCREASE THE NUMBER OF ROUND TRIP CONNECTIONS TO THE SPLUNK SERVER,
AND BLOCK THE CURRENT THREAD WHILE RETRYING.
```

fantavlik (approved):
Minor comment about calling out that this is blocking/synchronous but in general this looks good!

PR #447, opened 2022-04-14
Create job support for "output_mode:json"

ashah-splunk:
Added support for "output_mode='json'" while creating a job .


fantavlik (approved):
Nice, thanks for following up.

Issue #438, opened 2022-03-07
All Set-Cookie responses incorrectly assumed to be auth cookies


bendikro:
In short:
splunklib expects any Set-Cookie to be an auth cookie from Splunk. This is a problem when authenticating with a bearer token.

When hosting Splunk behind a load balancer like F5, which sets its own Set-Cookie in the response for connection persistence (stickiness), the Set-Cookie in the response from F5 will be included in the request headers instead of the splunk bearer token due to the logic in [splunklib/binding.py#L520](https://github.com/splunk/splunk-sdk-python/blob/ef88e9d3e90ab9d6cf48cf940c7376400ed759b8/splunklib/binding.py#L520)

When authenticating with user/pass, this makes sense, since Splunk API responds with a Set-Cookie for the session token (`splunkd_8089`). With a bearer token, Splunk does not respond with a session token, however the [_auth_headers](https://github.com/splunk/splunk-sdk-python/blob/ef88e9d3e90ab9d6cf48cf940c7376400ed759b8/splunklib/binding.py#L511) function seems to expect any Set-Cookie to be a session token.

The result is that, with bearer token authentication, requests to the Splunk API after the initial request will fail with "Request failed: Session is not logged in"

**To Reproduce**
1) Host Splunk behind F5 with stickiness enabled (Or any middleware that adds a Set-Cookie to the HTTP response)
2) Create a `splunklib.client.Service` with bearer token authentication
3) Perform a call that requires authentication and see that it fails

```
client = splunklib.client.Service(host='..', splunkToken='XXXX')
client.info()
```

**Expected behavior**
Authentication should work

**Workaround**
By passing the splunk token header manually, splunklib will include both the stickiness cookie as well as the splunk token in the requests.

```
client = splunklib.client.Service(host='..', splunkToken='XXXX', headers=[('Authorization', 'Splunk {}'.format('<token>'))])
client.info()
```

**Splunk:**
- Version: Any, presumably

**SDK:**
 - Version: Tested with 1.6.16

akaila-splunk:
Thanks for reporting this issue with the detailed information, we'll investigate it and try to provide a fix.

ashah-splunk:
Hi @bendikro , we have added the fix and it will be available in the next release. [Reference](https://github.com/splunk/splunk-sdk-python/pull/449/files#diff-701b42a0245cf832be44eb595d99bef52b3afca77ae631caa24643ed1ea6c4f3)

[...]

PR #449, opened 2022-04-22
Set cookie issue 438

ashah-splunk:
Changes in response to the issue [#438](https://github.com/splunk/splunk-sdk-python/issues/438)
- Updated the hasCookie method to check for the cookie received from Splunk (i.e "splunkd_<port>") instead of the existing length check.
- Updated the test cases with respect to the code changes.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -499,13 +499,13 @@ class Context(object):
         return self.http._cookies
 
     def has_cookies(self):
-        """Returns true if the ``HttpLib`` member of this instance has at least
-        one cookie stored.
+        """Returns true if the ``HttpLib`` member of this instance has auth token stored.
 
-        :return: ``True`` if there is at least one cookie, else ``False``
+        :return: ``True`` if there is auth token present, else ``False``
         :rtype: ``bool``
         """
-        return len(self.get_cookies()) > 0
+        auth_token_key = "splunkd_8089"
+        return auth_token_key in self.get_cookies().keys()
 
     # Shared per-context request headers
     @property
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -518,23 +518,23 @@ class Context(object):
 
         :returns: A list of 2-tuples containing key and value
         """
-        if self.has_cookies():
-            return [("Cookie", _make_cookie_header(list(self.get_cookies().items())))]
-        elif self.basic and (self.username and self.password):
-            token = 'Basic %s' % b64encode(("%s:%s" % (self.username, self.password)).encode('utf-8')).decode('ascii')
-            return [("Authorization", token)]
-        elif self.bearerToken:
-            token = 'Bearer %s' % self.bearerToken
-            return [("Authorization", token)]
-        elif self.token is _NoAuthenticationToken:
-            return []
-        else:
-            # Ensure the token is properly formatted
+        if self.token is not _NoAuthenticationToken:
             if self.token.startswith('Splunk '):
                 token = self.token
             else:
                 token = 'Splunk %s' % self.token
             return [("Authorization", token)]
+        elif self.bearerToken:
+            token = 'Bearer %s' % self.bearerToken
+            return [("Authorization", token)]
+        elif self.basic and (self.username and self.password):
+            token = 'Basic %s' % b64encode(("%s:%s" % (self.username, self.password)).encode('utf-8')).decode('ascii')
+            return [("Authorization", token)]
+        elif self.has_cookies():
+            return [("Cookie", _make_cookie_header(list(self.get_cookies().items())))]
+        else:
+            # no Authentication token
+            return []
 
     def connect(self):
         """Returns an open connection (socket) to the Splunk instance.
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -504,8 +504,8 @@ class Context(object):
         :return: ``True`` if there is auth token present, else ``False``
         :rtype: ``bool``
         """
-        auth_token_key = "splunkd_8089"
-        return auth_token_key in self.get_cookies().keys()
+        auth_token_key = "splunkd_"
+        return any(auth_token_key in key for key in self.get_cookies().keys())
 
     # Shared per-context request headers
     @property
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -518,23 +518,23 @@ class Context(object):
 
         :returns: A list of 2-tuples containing key and value
         """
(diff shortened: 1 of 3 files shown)


bendikro:
This fix now ensures that cookies only will be used when there is an auth cookie `splunkd_<XXXX>` availlable in `self.http._cookies`. However, it introduces a new bug which is that if there is a cookie available, e.g. from a Set-Cookie from a middleware, when there is not an auth cookie availlable, this cookie will not be included in the requests.

fantavlik:
> This fix now ensures that cookies only will be used when there is an auth cookie `splunkd_<XXXX>` availlable in `self.http._cookies`. However, it introduces a new bug which is that if there is a cookie available, e.g. from a Set-Cookie from a middleware, when there is not an auth cookie availlable, this cookie will not be included in the requests.

Hi @bendikro - the fact that any/all cookies were being sent if detected was an unintentional side-effect of the bug that was addressed in this change. By design the SDK should not include cookies unrelated to Splunk - although any additional headers including cookies can be set by users to be sent by the SDK client here: https://github.com/splunk/splunk-sdk-python/blob/19acb9a3944d09f52374b7c9fb829ccff277dc22/splunklib/binding.py#L454-L455 does this address your concern here? I hope I'm capturing your middleware use case.

bendikro:
> Hi @bendikro - the fact that any/all cookies were being sent if detected was an unintentional side-effect of the bug that was addressed in this change. By design the SDK should not include cookies unrelated to Splunk - although any additional headers including cookies can be set by users to be sent by the SDK client here:
>
> https://github.com/splunk/splunk-sdk-python/blob/19acb9a3944d09f52374b7c9fb829ccff277dc22/splunklib/binding.py#L454-L455
> does this address your concern here? I hope I'm capturing your middleware use case.

Hi @fantavlik

What is the rationale for not including cookies in the requests when a Set-Cookie has been included in a response? What is gained by ignoring these?

With "cookies unrelated to Splunk" you mean Set-Cookies that are not set by the Splunk server?
I challenge your definition of "unrelated to Splunk", and for our case I'd argue the Set-Cookie (`BIGipServer<MyPoolName>`) from the [F5 load balancer](https://www.f5.com/services/resources/glossary/load-balancer) is very much _related_ to Splunk even though it's not the Splunk server that generates the Set-Cookie. And by ignoring this Set-Cookie (set by the F5 middleware in our case), the Splunk SDK doesn't work properly when the network is configured with any load balancer that utilizes this technique.

The official Splunk docs even has a section on [Use a load balancer with search head clustering](https://docs.splunk.com/Documentation/Splunk/8.2.6/DistSearch/UseSHCwithloadbalancers):

```
Splunk recommends that you run a third-party hardware or software load balancer in front of your set of clustered search heads. That way, users can access the set of search heads through a single interface, without needing to specify a particular one.

There are a variety of third-party load balancers available that you can use for this purpose. Select a load balancer that employs layer-7 (application-level) processing.

Configure the load balancer so that user sessions are "sticky" or "persistent." This ensures that the user remains on a single search head throughout their session.
```

From this we find:
* Splunk recommends using a load balancer in front of the clustered search heads.
* The sticky sessions technique is specifically mentioned to ensure the session requests remain on a single search head.

The F5 doc has a page [HTTP Load Balancing with Cookie Persistence](https://techdocs.f5.com/en-us/bigip-14-0-0/big-ip-local-traffic-manager-implementations-14-0-0/configuring-http-load-balancing-with-cookie-persistence.html), where  "Cookie Persistence" is the "sticky" or "persistent" sessions referenced in the Splunk docs above.

The solution to include additional headers manually in the client is how we worked around the issue in the first place, but it shouldn't be necessary to manually extract the BIGip cookie (after initial requests to Splunk) and then create a new client with this cookie as an additional header. It is the responsibility of the Splunk SDK client to handle this.

If this behavior of the SDK truly is by design, I think the design should be reconsidered.

fantavlik:
Hi @bendikro thanks for giving us a detailed breakdown of your use case. I believe we've addressed your request in this PR: https://github.com/splunk/splunk-sdk-python/pull/463 which will be part of our next release in the coming days.

Trek333:
@fantavlik and @bendikro  The latest version of the [SOAR Splunk app](https://splunkbase.splunk.com/app/5848/) is impacted by this issue also. I would suggest that the next release of the Splunk Enterprise Python SDK features list should mention support of Load Balancer "sticky sessions" (persistent cookies) to make it more clear which software package that handles this feature.

The Splunk Enterprise Python SDK is included with the SOAR Splunk app to connect to a Splunk Enterprise Search Head (SH) server and/or Search Head Cluster (SHC). Using the proxy feature of the Burp Suite app to observe the behavior of the http requests & responses by the SOAR Splunk app via the Splunk Enterprise Python SDK to a Splunk Enterprise SHC with a F5 Load Balancer in front of it, the BIGipServer<MyPoolName> cookie used for persistence ("sticking sessions") is not included in subsequent http requests after receiving the BIGipServer<MyPoolName> cookie in the response of the first http request from the Splunk SOAR app. However as a comparison using a Firefox browser instead of the SOAR Splunk app as client to the Splunk Enterprise SHC, a Firefox browser does send the BIGipServer<MyPoolName> cookie in subsequent http requests after receiving the BIGipServer<MyPoolName> cookie in the response of the first http request from the Firefox browser.

bendikro:
@Trek333 I agree the feature should be listed, and the unit tests should be updated with a test to avoid any future regressions.

ashah-splunk:
@bendikro we did include unit tests where if the Auth Cookie is not available, authentication is being done with username and password. Test cases were updated as part of this PR itself, would request you to refer them. Also there were multiple PRs for the fix of the issue #438 and the fix was delivered in 2 different releases. As for mentioning this feature it was mentioned in the CHANGELOG.md but under **_Bug Fix(version 1.6.20)_** and **_Minor Changes(version 1.7.0)_** - [Reference](https://github.com/splunk/splunk-sdk-python/blob/master/CHANGELOG.md). I hope I have answered your concern, please do let me know if anything I have missed. Thanks

Trek333:
@ashah-splunk and @bendikro [#485 pull request](https://github.com/splunk/splunk-sdk-python/pull/485) has been submitted to add a unit test that is specific to regression testing the persistence of 3rd party cookies by the Splunk Enterprise Python SDK. More specifically, this unit test is targeted at the use case of the Splunk Enterprise Python SDK supporting "sticky sessions" of a Load Balancer for a Splunk Enterprise SHC (Search Head Cluster). Please let me know your feedback and feel to free to add more unit tests for 3rd party cookie persistence. Thanks

PR #460, opened 2022-06-01
Remove restart from client.Entity.disable

vmalaviya-splunk:
restart() call has been removed from the client.Entity.disable method


vishwambhar (approved):

vmalaviya-splunk:
> Interesting, so no unit test case impacted by this change?

No @vishwambhar, All the tests passed for the change

PR #463, opened 2022-06-08
Preserve thirdparty cookies

ashah-splunk:
- added cookies into the request header if thirdparty cookies exisits

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -543,7 +543,7 @@ class Context(object):
                 token = 'Splunk %s' % self.token
         if token:
             header.append(("Authorization", token))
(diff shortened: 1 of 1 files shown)


fantavlik (approved):

bendikro:
Were the tests updated to ensure there won't be any regressions?

ashah-splunk:
@bendikro yes tests were updated in PR #449 as part of the fix for issue #438. As we cannot have a setup with a middleware on the CI test execution we do not have tests for the exact scenario instead we have test cases where some cookie exists but not the Splunk Auth Cookie and authentication is carried out using username and password. Here Username/Password can be substituted by either Session key or bearer token.

PR #465, opened 2022-06-10
UnicodeDecodeError caught for json_loads

john-corcoran:
Hi, I encountered a `UnicodeDecodeError` when reading some data that was quite badly corrupted. This occurred when iterating through a `JSONResultsReader`. Error occurred using Python 3.10.5.

Error was:

`UnicodeDecodeError: 'utf-8' codec can't decode byte [val] in position [val]: invalid continuation byte`

The stacktrace points to `json.loads` (imported as `json_loads`) - this pull request catches the exception and re-reads, ignoring the decode error. It may be beneficial to raise a warning to the user, but apologies as I'm not familiar enough with the codebase for that just yet :)

### DIFF
--- a/splunklib/results.py
+++ b/splunklib/results.py
@@ -358,7 +358,12 @@ class JSONResultsReader(object):
         for line in stream.readlines():
             strip_line = line.strip()
             if strip_line.__len__() == 0: continue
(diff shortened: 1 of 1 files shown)


Ickerday:

PR #472, opened 2022-07-08
add prd support for eventing command

zhangziyi96:
add prd support for eventing command. Add partition_fields in ConfigurationSetting which makes custom set partition keys by themselves


Ickerday:

PR #473, opened 2022-07-18
Add decorator to simplify creating JSON REST endpoints

marcusschiesser:
Adds the decorator `json_handler` to simplify creating JSON REST endpoints. It automatically adds an exception handler and injects the following parameters to the `handle` method:

1. `request`, which is `json.loads(in_string)`
2. `query`, which contains the URL parameters as a `dict`
3. `payload`, which contains the payload as `dict` (remember to set `passPayload = true` in `restmap.conf`)

Here's an example for a GET handler with URL and path parameters:

```python
from splunk.persistconn.application import PersistentServerConnectionApplication
from splunklib.customrest import json_handler

class GetHandler(PersistentServerConnectionApplication):
    def __init__(self, _command_line, _command_arg):
        PersistentServerConnectionApplication.__init__(self)

    @json_handler
    def handle(self, request, query, **kwargs):
        path_param = request['path_info']
        url_param1 = query.get('param1', 1)
        url_param2 = query.get('param2', 1)
        return {'payload': {'path_param': path_param, 'url_param1': url_param1, 'url_param2': url_param2},
                'status': 200
                }
```

(Called with something like: http://localhost:8000/en-US/splunkd/__raw/servicesNS/-/{app-name}/myapi/333?param1=4)

And an example for a POST handler (with `passPayload = true`) which processes a JSON payload:

```python
from splunk.persistconn.application import PersistentServerConnectionApplication
from splunklib.customrest import json_handler

class PostHandler(PersistentServerConnectionApplication):
    def __init__(self, _command_line, _command_arg):
        PersistentServerConnectionApplication.__init__(self)

    @json_handler
    def handle(self, payload, **_kwargs):
        data = payload['flag']
        
        return {'payload': {'data': data},
                'status': 200
```                
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -140,7 +140,8 @@ setup(
 
     packages = ["splunklib",
                 "splunklib.modularinput",
-                "splunklib.searchcommands"],
+                "splunklib.searchcommands",
+                "splunklib.customrest"],
 
     url="http://github.com/splunk/splunk-sdk-python",
 
--- a/splunklib/customrest/__init__.py
+++ b/splunklib/customrest/__init__.py
@@ -1,0 +1,6 @@
+"""The following imports allow these classes to be imported via
+the splunklib.customrest package like so:
+
+from splunklib.customrest import *
+"""
+from .json import json_handler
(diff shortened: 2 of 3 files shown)


Ickerday:

PR #474, opened 2022-07-27
add prd for eventing and reporting command

zhangziyi96:
add prd support for custom command


Ickerday:

PR #491, opened 2022-10-06
The correct permissions on event_writer.py are 644 or -rw-r--r--, it …

gjanders:
…currently has execute permission and this results in a failure to pass appinspect


ashah-splunk:
@gjanders closing your PR as we have a new PR with the file permission updated as well a new test case to check for the file permission to prevent such issue in future. Ref - #493

Issue #497, opened 2022-11-18
KVStoreCollection using wrong argument in "update_index"


seiimonn:
**Describe the bug**
When updating the index of a collection the following error is given:

`splunklib.binding.HTTPError: HTTP 400 Bad Request -- Argument "index.ho1" is not supported by this handler.`

**To Reproduce**
```python
>>> import splunklib.client as client
>>> c = client.connect()
>>> c.kvstore.create("example_collection")
>>> c.kvstore["example_collection"].update_index("get_fast",{"important_field": 1})
```

**Expected behavior**
Update of accelerated fields for collection.

**Logs or Screenshots**
```
>>> c.kvstore["example_collection"].update_index("get_fast",{"important_field": 1})
/usr/lib/python3/dist-packages/requests/__init__.py:89: RequestsDependencyWarning: urllib3 (1.26.11) or chardet (3.0.4) doesn't match a supported version!
  warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported "
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/app/backend/src/lib/splunklib/client.py", line 3622, in update_index
    return self.post(**kwargs)
  File "/app/backend/src/lib/splunklib/client.py", line 1047, in post
    return super(Entity, self).post(path_segment, owner=owner, app=app, sharing=sharing, **query)
  File "/app/backend/src/lib/splunklib/client.py", line 852, in post
    return self.service.post(path, owner=owner, app=app, sharing=sharing, **query)
  File "/app/backend/src/lib/splunklib/binding.py", line 289, in wrapper
    return request_fun(self, *args, **kwargs)
  File "/app/backend/src/lib/splunklib/binding.py", line 70, in new_f
    val = f(*args, **kwargs)
  File "/app/backend/src/lib/splunklib/binding.py", line 769, in post
    response = self.http.post(path, all_headers, **query)
  File "/app/backend/src/lib/splunklib/binding.py", line 1262, in post
    return self.request(url, message)
  File "/app/backend/src/lib/splunklib/binding.py", line 1282, in request
    raise HTTPError(response)
splunklib.binding.HTTPError: HTTP 400 Bad Request -- Argument "index.get_fast" is not supported by this handler.
```

**Splunk (please complete the following information):**
- Version: 9.0.2
- OS: Ubuntu 20.04 LTS
- Deployment: tested on single-instance

**SDK (please complete the following information):**
 - Version: 1.6.19
[...]

PR #500, opened 2022-12-16
update support for accelerated_field of kvstore

akaila-splunk:
updated accelerated_field methods and added test cases for it. (issue #497 for ref)

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -3748,7 +3748,7 @@ class KVStoreCollection(Entity):
         return KVStoreCollectionData(self)
 
     def update_accelerated_field(self, name, value):
-        """Changes the definition of a KV Store index.
+        """Changes the definition of a KV Store accelerated_field.
 
         :param name: name of accelerated_fields to change
         :type name: ``string``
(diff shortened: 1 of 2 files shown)


splunklib/client.py
```diff
@@ -3740,18 +3747,20 @@ def data(self):
         """
         return KVStoreCollectionData(self)
 
-    def update_index(self, name, value):
```
fantavlik, line 3743: I think we want to add support for accelerated field indexes without removing current index support

fantavlik (changes requested):
~Recommend adding accelerated fields support without removing current functionality if possible~

Since this functionality was deprecated in 7.x I'm good with moving forward.

splunklib/client.py
akaila-splunk: In Splunk Enterprise v6.4 archive docs we found that for kvstore `index.<indexName>` property exist but in later version it was updated to `accelerated_field.<field_name>`, so due to this change we updated the same method. PFA docs screenshots for reference, let us know if we still want existing index support.

Splunk v6.4 docs:
splunkv6 4_doc

Splunk v8.2 docs:
splunkv8 2_doc

vishwambhar (approved):

splunklib/client.py
fantavlik: OK great thanks for clarifying, in that case the changes look good to me 👍

Issue #506, opened 2023-01-12
SDK logs sensitive data when running in debug mode


pablogrigo:
**Describe the bug**
SDK logs sensitive data when running in debug mode.

**To Reproduce**
1. Configure logging in debug mode. (https://github.com/splunk/splunk-sdk-python#optionalset-up-logging-for-splunklib)
2. Call some specific endpoints (see list at the bottom).
3. Inspect the logs, note passwords are included in plain text.

**Expected behaviour**
No sensitive data is logged, even when running in debug mode.

**Logs or Screenshots**
The `binding.py` file contains several instances of logging of network operations.
Example 1: https://github.com/splunk/splunk-sdk-python/blob/master/splunklib/binding.py#L779
Example 2: https://github.com/splunk/splunk-sdk-python/blob/master/splunklib/binding.py#L845

**Splunk**
- Version: Any
- OS: Any
- Deployment: Any

**SDK:**
 - Version: 1.7.2
 - Language Python 3.X
 - OS: Any

**Additional context**
There was an attempt to prevent password leaks, as seen in #439 
Unfortunately, according to the API Reference https://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTlist there are several endpoints that will include either a password or a key in their body, to list a few:

- [auth/login](http://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTaccess#auth.2Flogin)
- [authentication/users](http://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTaccess#authentication.2Fusers)
- [authentication/users/{name}](http://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTaccess#authentication.2Fusers.2F.7Bname.7D)
- [authentication/providers/SAML](http://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTaccess#authentication.2Fproviders.2FSAML)
- [authentication/providers/SAML/{stanza_name}](http://docs.splunk.com/Documentation/Splunk/9.0.3/RESTREF/RESTaccess#authentication.2Fproviders.2FSAML.2F.7Bstanza_name.7D)

The recommendation would be to avoid logging headers and payloads of network operations, as sensitive data could be included in both.

ashah-splunk:
[...]

PR #507, opened 2023-01-20
masked sensitive data in logs

akaila-splunk:
added new method to mask sensitive data in logs like password. changes wrt issue #506

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -61,7 +61,10 @@ __all__ = [
     "HTTPError"
 ]
 
-SENSITIVE_KEYS = ["password", "token", "Authorization"]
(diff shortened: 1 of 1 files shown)


splunklib/binding.py
```diff
@@ -60,12 +61,14 @@
     "HTTPError"
 ]
 
+SENSITIVE_KEYS = ["password", "token", "Authorization"]
```
l00py, line 64: May want to review the request params for the supported endpoints specified here: https://github.com/splunk/splunk-sdk-python/blob/develop/splunklib/client.py#L90-L113

For example:

- apps/local
  - auth
  - session
- configs/conf
  - POST to conf files may contain sensitive information, but we can do a best-effort based on the current conf specs, and what files are normally interacted with. It may be best to provide a different logic for masking sensitive information for that specific endpoint.

l00py (commented):
Added comments for review

PR #526, opened 2023-04-19
Marked newly added test cases for smoke test

akaila-splunk:
Marked newly added test cases (~in last 2 year) of CSC app, cookies and searches for smoke test


fantavlik (approved):
Looks good, thanks for the investigation!

PR #545, opened 2023-09-24
Update to binding.py

pch3:
Although port number in HTTP Host header is not mandatory, it may be used by load balancer to route traffic to the right cluster.
Change is very simple and adds ":port" part to the Host HTTP header.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -1457,7 +1457,7 @@ def handler(key_file=None, cert_file=None, timeout=None, verify=False, context=N
         body = message.get("body", "")
         head = {
             "Content-Length": str(len(body)),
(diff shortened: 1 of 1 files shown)


splunklib/binding.py
```diff
@@ -1457,7 +1457,7 @@ def request(url, message, **kwargs):
         body = message.get("body", "")
         head = {
             "Content-Length": str(len(body)),
-            "Host": host,
+            "Host": "%s:%d" % (host, port),
```
Ickerday:
```suggestion
            "Host": f"{host}:{port}"
```

mateusz834:
Hi

I will close that in favor of #673

PR #546, opened 2023-09-26
Fossa scan

ashah-splunk:
- added FOSSA scan as part of CI testing


l00py (approved):
LGTM
1. `six.py` will be removed in a future release
2. Future action item: for 3rd party library, must use `requirements.txt` and `setup.py` and include at build time.
3. Fossa scan requirements: https://docs.fossa.com/docs/python

PR #559, opened 2024-02-22
feat: Add exception logging in Script

kkedziak-splunk:
This change adds logging of exceptions - exception type, value and traceback, to scripted modular input.

Unfortunately such logs cannot be multi-line due to limitations of [ExecProcessor](https://cd.splunkdev.com/splcore/main/-/blob/develop/src/input/ExecProcessor.cpp). Multi-line log is always interpreted as separate one-line logs.


sgoral-splunk (approved):

sgoral-splunk:
@kkedziak-splunk  we should change PR name

maszyk99 (approved):

Issue #548, opened 2023-10-03
Failed installing with `pip install splunk-sdk` for Python 3.12


jfcherng:
**Describe the bug**

Install this with Python 3.12.0 via pip failed.

**To Reproduce**
1. Execute `python -m pip install splunk-sdk`, where `python` is Python 3.12.0

**Expected behavior**

It installs successfully.

**Logs or Screenshots**

```
(.venv) [jfcherng@3bc-f19 test]$ pip install splunk-sdk
Collecting splunk-sdk
  Using cached splunk-sdk-1.7.4.tar.gz (108 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [22 lines of output]
      Traceback (most recent call last):
        File "D:\jfcherng\Desktop\test\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
          main()
        File "D:\jfcherng\Desktop\test\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
          json_out['return_val'] = hook(**hook_input['kwargs'])
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "D:\jfcherng\Desktop\test\.venv\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 118, in get_requires_for_build_wheel
          return hook(config_settings)
                 ^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\jfcherng\AppData\Local\Temp\pip-build-env-cu0raqgr\overlay\Lib\site-packages\setuptools\build_meta.py", line 355, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=['wheel'])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\jfcherng\AppData\Local\Temp\pip-build-env-cu0raqgr\overlay\Lib\site-packages\setuptools\build_meta.py", line 325, in _get_build_requires
          self.run_setup()
[...]

PR #569, opened 2024-03-22
Add "deprecation" package as a dependency

jfcherng:
Fixes https://github.com/splunk/splunk-sdk-python/issues/548#issuecomment-2014425756

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,6 +142,10 @@ setup(
                 "splunklib.modularinput",
                 "splunklib.searchcommands"],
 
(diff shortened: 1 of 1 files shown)


ashah-splunk:
Hi @jfcherng 
Thanks for the PR, as we needed to remove the `deprecation` dependency from the `tox.ini` as well. We created a new PR for the fix. Hence closing this PR. Also Splunk Python SDK v2.0.1 has been published with the fix

PR #575, opened 2024-05-15
fix: delete in binding.py now same as get/post

DanielPerezJensen:
Currently you can not pass custom headers to delete, as they are overwritten with default self._auth_headers. Replaced that with all_headers as in get and put.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -662,8 +662,9 @@ class Context:
         """
         path = self.authority + self._abspath(path_segment, owner=owner,
                                               app=app, sharing=sharing)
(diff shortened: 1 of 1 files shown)


DanielPerezJensen:
This change is required to leverage .delete using splunklib. Current work is blocked by this, as deletion is not possible without header authentication. Please have a look at this, or any other way we can expedite this.

DanielPerezJensen:
Saw in another PR wanting to solve the same thing, to add a test case. Where would you suggest I do that? In which testing file is this functionality tested usually?

Ickerday:

PR #586, opened 2024-10-07
Adding helper functions for convenient PUT and PATCH on SDK

seiimonn:
The current SDK does not provide dedicated functions for sending PUT and PATCH requests to endpoints in Splunk. These operations were only possible using the generic request method, which had issues correctly formatting a body payload, especially for form-encoded data.

This change introduces helper functions for PUT and POST requests, along with a new _prepare_request_body_and_url function to streamline body and URL preparation logic. This ensures consistency, reduces redundancy, and resolves issues with incorrect payload formatting when making these requests.

## Changes

- Added a _prepare_request_body_and_url helper function to handle body preparation and URL encoding for POST and PUT requests.
- Added put() and updated post() methods to leverage the helper function.
- Ensured proper handling of Content-Type headers for form-encoded data.

## Benefits

- Simplifies making PUT and PATCH requests directly without needing the generic request method.
- Fixes body formatting issues, ensuring payloads are correctly encoded.

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -809,6 +809,171 @@ class Context:
         response = self.http.post(path, all_headers, **query)
         return response
 
+
+    @_authentication
+    @_log_duration
+    def put(self, path_segment, object, owner=None, app=None, sharing=None, headers=None, **query):
+        """Performs a PUT operation from the REST path segment with the given object,
+        namespace and query.
+
+        This method is named to match the HTTP method. ``put`` makes at least
+        one round trip to the server, one additional round trip for each 303
+        status returned, and at most two additional round trips if
+        the ``autologin`` field of :func:`connect` is set to ``True``.
+
+        If *owner*, *app*, and *sharing* are omitted, this method uses the
+        default :class:`Context` namespace. All other keyword arguments are
+        included in the URL as query parameters.
+
+        Some of Splunk's endpoints, such as ``receivers/simple`` and
+        ``receivers/stream``, require unstructured data in the PUT body
+        and all metadata passed as GET-style arguments. If you provide
+        a ``body`` argument to ``put``, it will be used as the PUT
+        body, and all other keyword arguments will be passed as
+        GET-style arguments in the URL.
+
+        :raises AuthenticationError: Raised when the ``Context`` object is not
+             logged in.
+        :raises HTTPError: Raised when an error occurred in a GET operation from
+             *path_segment*.
+        :param path_segment: A REST path segment.
+        :type path_segment: ``string``
+        :param object: The object to be PUT.
+        :type object: ``string``
+        :param owner: The owner context of the namespace (optional).
+        :type owner: ``string``
+        :param app: The app context of the namespace (optional).
+        :type app: ``string``
+        :param sharing: The sharing mode of the namespace (optional).
+        :type sharing: ``string``
+        :param headers: List of extra HTTP headers to send (optional).
+        :type headers: ``list`` of 2-tuples.
+        :param query: All other keyword arguments, which are used as query
+            parameters.
+        :param body: Parameters to be used in the post body. If specified,
+            any parameters in the query will be applied to the URL instead of
+            the body. If a dict is supplied, the key-value pairs will be form
+            encoded. If a string is supplied, the body will be passed through
+            in the request unchanged.
+        :type body: ``dict`` or ``str``
+        :return: The response from the server.
+        :rtype: ``dict`` with keys ``body``, ``headers``, ``reason``,
+                and ``status``
+
+        **Example**::
+
+            c = binding.connect(...)
+            c.post('saved/searches', name='boris',
+                   search='search * earliest=-1m | head 1') == \\
+                {'body': ...a response reader object...,
+                 'headers': [('content-length', '10455'),
+                             ('expires', 'Fri, 30 Oct 1998 00:00:00 GMT'),
+                             ('server', 'Splunkd'),
+                             ('connection', 'close'),
+                             ('cache-control', 'no-store, max-age=0, must-revalidate, no-cache'),
+                             ('date', 'Fri, 11 May 2012 16:46:06 GMT'),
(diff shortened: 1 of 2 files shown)


Ickerday:

seiimonn:
This is still very much sought after by my team. Let me know if you need anything for this to be merged.

Ickerday:
Thanks @seiimonn, we'll take a look and get back to you!

seiimonn:
Hi @Ickerday - anything I can do to support on this?

mateusz834:
Hi

We plan to give this a look, we are going to figure out which way is the best to implement this (we might go with this change) but in the meantime can you rebase this on top of the develop branch?

mateusz834:
Thanks, once again i would ask you to rebase this, since i would like to see this being also tested with the tests i have introduced for #683.

Please add in `tests/system/test_cre_apps.py` `TestJSONCustomRestEndpointsSpecialMethodHelpers` a test case for PUT and PATCH, i assume they would look the same as `test_POST`.

splunklib/binding.py
```diff
@@ -942,8 +1107,9 @@ def request(
             body = _encode(**body)
 
             if method == "GET":
-                path = path + UrlEncoded("?" + body, skip_encode=True)
-                message = {"method": method, "headers": all_headers}
+                path = path + UrlEncoded('?' + body, skip_encode=True)
+                message = {'method': method,
+                           'headers': all_headers}
```
mateusz834, line 1112: Drop that change. It seems to only make the code non-formatted.

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
+                             ('date', 'Fri, 11 May 2012 16:46:06 GMT'),
+                             ('content-type', 'text/xml; charset=utf-8')],
+                 'reason': 'Created',
+                 'status': 201}
+            c.post('nonexistant/path') # raises HTTPError
+            c.logout()
+            # raises AuthenticationError:
+            c.put('saved/searches/boris',
+                   search='search * earliest=-1m | head 1')
+        """
```
mateusz834, line 936: I think we should change this example, splunk does not really support PUT/PATCH in its endpoints see: https://help.splunk.com/en/splunk-enterprise/leverage-rest-apis/rest-api-reference/10.0/introduction/endpoints-reference-list (that is the reason there were no such methods in the first place) 

I think we should have an example that shows that it is used with custom rest endpoints, something like:

```py
# Call an HTTP endpoint, exposed as Custom Rest Endpoint in a Splunk App.
# PUT /servicesNS/-/app_name/custom_rest_endpoint
service.put(
     app="app_name",
     path_segment="custom_rest_endpoint",
     body=json.dumps({"key": "val"}),
     headers=[("Content-Type", "application/json")],
)
```

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
+                             ('cache-control', 'no-store, max-age=0, must-revalidate, no-cache'),
+                             ('date', 'Fri, 11 May 2012 16:46:06 GMT'),
+                             ('content-type', 'text/xml; charset=utf-8')],
+                 'reason': 'Created',
+                 'status': 201}
+            c.post('nonexistant/path') # raises HTTPError
+            c.logout()
+            # raises AuthenticationError:
+            c.patch('saved/searches/boris',
+                   search='search * earliest=-1m | head 1')
```
mateusz834, line 1017: Same here, lets change it, as above.

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
+        :type owner: ``string``
+        :param app: The app context of the namespace (optional).
+        :type app: ``string``
+        :param sharing: The sharing mode of the namespace (optional).
+        :type sharing: ``string``
+        :param headers: List of extra HTTP headers to send (optional).
+        :type headers: ``list`` of 2-tuples.
+        :param query: All other keyword arguments, which are used as query
+            parameters.
+        :param body: Parameters to be used in the post body. If specified,
```
mateusz834, line 906:
```suggestion
        :param body: Parameters to be used in the put body. If specified,
```

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
```
mateusz834, line 988:
```suggestion
        :param body: Parameters to be used in the patch body. If specified,
```

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
+        If *owner*, *app*, and *sharing* are omitted, this method uses the
+        default :class:`Context` namespace. All other keyword arguments are
+        included in the URL as query parameters.
+
+        Some of Splunk's endpoints, such as ``receivers/simple`` and
+        ``receivers/stream``, require unstructured data in the PUT body
+        and all metadata passed as GET-style arguments. If you provide
+        a ``body`` argument to ``put``, it will be used as the PUT
+        body, and all other keyword arguments will be passed as
+        GET-style arguments in the URL.
```
mateusz834, line 886: Splunk does not support PUT method in its endpoints, so lets not mention that.

```suggestion
        If you provide a ``body`` argument to ``put``, 
        it will be used as the PUT body, and all other 
        keyword arguments will be passed as GET-style 
        arguments in the URL.
```

splunklib/binding.py
```diff
@@ -862,6 +862,171 @@ def post(
...
+        If *owner*, *app*, and *sharing* are omitted, this method uses the
+        default :class:`Context` namespace. All other keyword arguments are
+        included in the URL as query parameters.
+
+        Some of Splunk's endpoints, such as ``receivers/simple`` and
+        ``receivers/stream``, require unstructured data in the PATCH body
+        and all metadata passed as GET-style arguments. If you provide
+        a ``body`` argument to ``patch``, it will be used as the PATCH
+        body, and all other keyword arguments will be passed as
+        GET-style arguments in the URL.
```
mateusz834, line 968: Same here.

PR #602, opened 2025-02-10
Use relative imports for splunk-sdk

jonathan-s:
Referencing https://github.com/splunk/splunk-sdk-python/pull/163 where there was an effort to use relative imports. There are still some spots where there are not relative imports. 

Using relative imports also makes it easier to vendor splunklib.

PR #613, opened 2025-04-09
Revert "Use relative imports for splunk-sdk"

szymonjas:
Reverts splunk/splunk-sdk-python#602

we shouldn't  directly merge to master


jonathan-recordedfuture:
@szymonjas, I'm curious why this was reverted, and it seems like this didn't make it into the library in the end.

PR #616, opened 2025-05-22
client.py-invalid-esc-seq

onequbit:
trying to address the invalid escape sequence error


mateusz834:
This error was fixed in: 71c984c024d1001124560410a68ac509d74086c4

Also See #618

Issue #582, opened 2024-09-04
SyntaxWarning: invalid escape sequence in Python 3.12


AlexPetul:
**Describe the bug**
When i updated my project to use Python 3.12 i started to get a warning from splunk-sdk:
/usr/local/lib/python3.12/site-packages/splunklib/client.py:772: SyntaxWarning: invalid escape sequence '\\/'

In Python 3.12 there was a change related to this issue:

> A backslash-character pair that is not a valid escape sequence now generates a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning), instead of [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). For example, re.compile("\d+\.\d+") now emits a [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning) ("\d" is an invalid escape sequence, use raw strings for regular expression: re.compile(r"\d+\.\d+")). In a future Python version, [SyntaxError](https://docs.python.org/3/library/exceptions.html#SyntaxError) will eventually be raised, instead of [SyntaxWarning](https://docs.python.org/3/library/exceptions.html#SyntaxWarning.)

**To Reproduce**
1. Just use Python3.12 with splunk-sdk

**SDK (please complete the following information):**
 - Version: 2.0.2
 - Language Runtime Version: Python 3.12
 - OS: Any

Ickerday:
Dismissed as stale.

PR #618, opened 2025-06-05
Added "raw-string-notation" character to regex string to fix bug caused by invalid escape sequence

chunks-hurry-0z:
Bugfix for #582 in accordance with the [Python Docs](https://docs.python.org/3/library/re.html#:~:text=Regular%20expressions%20use,rawstring%20notation.) re: "raw-string-notation".

See also alternative fix #616

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,7 @@
 name: Python CI
 
 on:
-  [ push, workflow_dispatch ]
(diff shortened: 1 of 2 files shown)


Synse (approved):

Ickerday:
Hi, sorry for the big wait - we're taking a look!

mateusz834:
Closing as we have already fixed that on master: 71c984c024d1001124560410a68ac509d74086c4.

PR #629, opened 2025-07-31
[TEST] Bartek's Lab

Ickerday:
Just me trying to break and fix stuff


.github/workflows/test.yml
```diff
@@ -1,51 +1,47 @@
...
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
-
-      - name: Install tox
-        run: pip install tox
-
-      - name: Test Execution
+      - name: Install dependencies
+        run: pip install .
```
cecyliaborek, line 41: this will install all dependencies defined in `pyproject` which I think is an overkill if we only need tox

pyproject.toml
```diff
@@ -0,0 +1,50 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "splunk-sdk"
+version = "2.1.0"
+description = "The Splunk Software Development Kit for Python"
+readme = "README.md"
+requires-python = ">=3.9, <4"
```
cecyliaborek, line 10: `<4` - why?

splunk-sdk-python.code-workspace
```diff
@@ -0,0 +1,53 @@
...
+      "mikestead.dotenv",
+      "ms-python.debugpy",
+      "ms-python.python",
+      "ms-python.vscode-pylance",
+      "ms-python.vscode-python-envs",
+      "ms-vscode.makefile-tools",
+      "streetsidesoftware.code-spell-checker",
+      "pflannery.vscode-versionlens",
+      "splunk.splunk",
+      "charliermarsh.ruff"
```
cecyliaborek, line 50: shouldn't for example ruff be added as dependency in `pyproject`

cecyliaborek (commented):
Leaving some comments/thoughts

README.md
```diff
@@ -60,45 +59,51 @@ Install the sources you cloned from GitHub:
...
-make wait_up
-make test
-make down
-```
+    ```shell
+    make up SPLUNK_VERSION=latest # also available: 10, 9.4
+    make wait_up
+    make test
+    make down
+    ```
```
mateusz834, line 67: Mention `cp .env.template .env`? Otherwise tests will fail with an `AuthenticationError`.

.github/workflows/test.yml
Ickerday: Yeah, in the end I want to [extract everything](https://github.com/splunk/splunk-sdk-python/pull/629/commits/7bd53fcca47c01d92736da9d40b667bc042fdae3) to an [optional group](https://devsjc.github.io/blog/20240627-the-complete-guide-to-pyproject-toml/#optional-dependencies) but I had issues with it (`WARNING: splunk-sdk 2.1.0 does not provide the extra 'dev'`) so I defaulted to installing everything instead while I do my other research.

pyproject.toml
Ickerday: I saw this being used as a convention sometime ago, but it's seemingly removed from official examples now

.python-version
```diff
@@ -0,0 +1 @@
+3.9.23
```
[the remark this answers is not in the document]
Ickerday: I'm defaulting to 3.9 as this is the version offered by both Splunk 9.* and 10.*. We have no way of installing 3.7 on Apple Silicon and I just can't be bothered to care about 3.7 for now tbh

PR #631, opened 2025-08-07
DVPL-12497: Divide tests into folders

cecyliaborek:
Divide tests into folders based on test types (unit/integration/system). 

Following PRs will have separate setup and run commands for different test types. This is needed because unit tests don't require a running splunk instance, integration and system require different setups of a Splunk instance.


cecyliaborek:
Closing, and opening #637 instead

PR #633, opened 2025-08-08
DVPL-12497: Fix CSC apps' tests

cecyliaborek:
Fixed a flaky test `tests/searchcommands/test_csc_apps.py::TestCSC::test_eventing_app` caused by:
- Unreliable data source – The test relied on events from Splunk’s `_internal` index, which is not guaranteed to contain the required data.
- Malformed search query - missing `status` field caused `eventingcsc status=200` to raise `KeyError`, which wasn’t caught by the test assertions.

**The Fix**
- Created a helper method `_create_test_data_search` that creates synthetic events for the `eventingcsc` command
- Made assertions stricter to check both result count and field values.

### DIFF
--- a/tests/searchcommands/test_csc_apps.py
+++ b/tests/searchcommands/test_csc_apps.py
@@ -22,16 +22,17 @@ from splunklib import results
 
 
 @pytest.mark.smoke
-class TestCSC(testlib.SDKTestCase):
-    def test_eventing_app(self):
-        app_name = "eventing_app"
+class TestEventingApp(testlib.SDKTestCase):
+    app_name = "eventing_app"
 
+    def test_metadata(self):
         self.assertTrue(
-            app_name in self.service.apps, msg="%s is not installed." % app_name
+            TestEventingApp.app_name in self.service.apps,
+            msg=f"{TestEventingApp.app_name} is not installed.",
         )
 
         # Fetch the app
-        app = self.service.apps[app_name]
+        app = self.service.apps[TestEventingApp.app_name]
         app.refresh()
 
         # Extract app info
@@ -40,6 +41,8 @@ class TestCSC(testlib.SDKTestCase):
         state = app.state
 
         # App info assertions
+        self.assertEqual(state.title, TestEventingApp.app_name)
+
         self.assertEqual(access.app, "system")
         self.assertEqual(access.can_change_perms, "1")
         self.assertEqual(access.can_list, "1")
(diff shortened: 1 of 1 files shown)


tests/searchcommands/test_csc_apps.py
```diff
@@ -64,17 +64,86 @@ def test_eventing_app(self):
...
+            msg for msg in informational_messages if msg.type in ['FATAL', 'ERROR']
+        ]
+        self.assertEqual(
+            len(fatal_messages),
+            0,
+            f"Should not have FATAL/ERROR messages, but got: {[msg.message for msg in fatal_messages]}",
+        )
+
+        first_result = actual_results[0]
+        self.assertIn('status', first_result, "Result should contain 'status' field")
```
Ickerday, line 93: These descriptions are redundant, let's get rid of them

Ickerday:
For reasons yet unknown to me, we check for `is_preview` in results - here's how I dealt with this for Splunk 10. Maybe it'll be useful
https://github.com/splunk/splunk-sdk-python/pull/629/commits/c14f4f388297a5da359a5a5ab5049221fc062aca

Ickerday:
To be honest, I don't understand what is going on here. Why are we testing for errors? Why are old tests removed?

tests/searchcommands/test_csc_apps.py
```diff
@@ -64,17 +64,39 @@ def test_eventing_app(self):
...
+
+        reader = results.JSONResultsReader(stream)
+        items = list(reader)
+
+        actual_results = [item for item in items if isinstance(item, dict)]
+        informational_messages = [
+            item for item in items if isinstance(item, results.Message)
+        ]
 
-        self.assertEqual(result.is_preview, False)
```
Ickerday, line 74: Why was this removed?

tests/searchcommands/test_csc_apps.py
```diff
@@ -64,17 +64,39 @@ def test_eventing_app(self):
...
+        ]
+        self.assertEqual(
+            len(fatal_messages),
+            0,
+            f"Should not have FATAL/ERROR messages, but got: {[msg.message for msg in fatal_messages]}",
+        )
+
+        first_result = actual_results[0]
+        self.assertIn('status', first_result)
+        self.assertEqual(first_result['status'], '200')
```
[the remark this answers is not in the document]
cecyliaborek: good point 👍

tests/searchcommands/test_csc_apps.py
Ickerday: I didn't ask to include it - I was curious why this was removed.

cecyliaborek: At first, I thought it wasn’t included in the response in Splunk 10. But it turned out it was - I guess it just wasn’t there when the search query was failing. This is the response I was getting for the malformed query:
` {'messages': [{'type': 'FATAL', 'text': 'Error in \'eventingcsc\' command: KeyError at "/opt/splunk/etc/apps/eventing_app/bin/eventingcsc.py", line 44 : \'status\''}]}`

Ickerday (approved):

PR #638, opened 2025-08-18
Mark non-test classes with __test__ = False attribute

mateusz834:
This resolves few warnings that were emitted by pytest.

See: https://docs.pytest.org/en/stable/example/pythoncollection.html#customizing-test-collection

### DIFF
--- a/tests/searchcommands/test_search_command.py
+++ b/tests/searchcommands/test_search_command.py
@@ -55,6 +55,8 @@ def build_command_input(getinfo_metadata, execute_metadata, execute_body):
 
 @Configuration()
 class TestCommand(SearchCommand):
+    __test__ = False
+
     required_option_1 = Option(require=True)
     required_option_2 = Option(require=True)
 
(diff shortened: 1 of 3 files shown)


tests/searchcommands/test_search_command.py
```diff
@@ -99,6 +101,8 @@ class ConfigurationSettings(SearchCommand.ConfigurationSettings):
 
 @Configuration()
 class TestStreamingCommand(StreamingCommand):
+    __test__ = False
+
```
cecyliaborek, line 105: this one is not being referenced anywhere, wondering if we could remove it (could always bring it back from the version control if needed)

cecyliaborek (commented):
lgtm 👍 
just wondering, any reason why you chose to use `__test__ = False` instead of renaming those non-test classes? e.g. `SearchCommandFixture` instead of `TestSearchCommand`. That could be a better self-documentation of the code

mateusz834:
Closing in favor of #667.

PR #642, opened 2025-08-21
Re-enable test_delete from test_input.py

mateusz834:
This also simplifies the test and splits it into to two separate tests.

### DIFF
--- a/tests/integration/test_input.py
+++ b/tests/integration/test_input.py
@@ -282,27 +282,35 @@ class TestInput(testlib.SDKTestCase):
             entity.refresh()
             self.assertEqual(entity.host, kwargs["host"])
 
-    @pytest.mark.skip("flaky")
     def test_delete(self):
-        inputs = self.service.inputs
-        remaining = len(self._test_entities) - 1
-        for input_entity in self._test_entities.values():
-            name = input_entity.name
-            kind = input_entity.kind
-            self.assertTrue(name in inputs)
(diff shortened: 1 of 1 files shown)


tests/integration/test_input.py
```diff
@@ -282,27 +282,30 @@ def test_update(self):
...
+        udpName, udpKind = (
+            self._test_entities["udp"].name,
+            self._test_entities["udp"].kind,
+        )
+        tcpName, tcpKind = (
+            self._test_entities["tcp"].name,
+            self._test_entities["tcp"].kind,
+        )
+
+        self._test_entities["tcp"].delete()
```
cecyliaborek, line 295: what is the reason for this? what should it do?

mateusz834: Basically this should behave in the same way as `self.service.inputs.delete` below does (i.e it removes the input created in the `setUp` from splunk).

mateusz834: > what is the reason for this?

To simultaneously test both ways of removing.

cecyliaborek: makes sense! I would maybe consider adding a small comment clarifying that, as at first glance it might seem confusing (e.g. `#Deleting via the Input entity itself` and `#Deleting via the Inputs collection`).

Otherwise the test is way better and clearer now 👍

cecyliaborek (approved):
One small suggestion, but do whatever you see fit

PR #643, opened 2025-08-21
DVPL-12468 separate tox targets

cecyliaborek:
- Update tests README with instructions
- Small improvements in tox.ini:
  - by default only search `tests` folder for tests
  - don't generate docs when running `tox` without specified env


Ickerday:
Closing as we move to remove `tox`.

PR #660, opened 2025-09-01
Modernize project to use `pyproject.toml`, review .gitignore and untrack any ignored files, update README

### LINKED ISSUE #620: Deprecation warning when installing due to legacy setup.py bdist_wheel mechanism used
**Describe the bug**
This project is using the deprecated setup.py bdist_wheel mechanism for installation which is going to lose support in a future version of pip. A pyproject.toml file at the root of the project will remedy this issue.

**To Reproduce**
1. Download source
2. In terminal, pip install .
3. Notice deprecation warning

**Expected behavior**
No deprecation warning

**Logs or Screenshots**
````
DEPRECATION: Building 'splunk-sdk' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the '--use-pep517' option, (possibly combined with '--no-build-isolation'), or adding a 'pyproject.toml' file to the source tree of 'splunk-sdk'. Discussion can be found at https://github.com/pypa/pip/issues/6334
````

 - Version: latest
 - Language Runtime Version: Python 3.9
 - OS: Windows 10

**Additional context**
To remedy this a pyproject.toml file should be added at the root of the project with the following:
````
# pyproject.toml
[build-system]
# XXX: If your project needs other packages to build properly, add them to this list.
requires = ["setuptools >= 42.0.0"]
build-backend = "setuptools.build_meta"
````

More information can be found here: https://github.com/pypa/pip/issues/6334

### PR DESCRIPTION
**Pull request by Ickerday**
Closes #620 

```sh
❯ twine check dist/*
Checking dist/splunk_sdk-2.1.1-py3-none-any.whl: PASSED
Checking dist/splunk_sdk-2.1.1.tar.gz: PASSED
```


pyproject.toml
```diff
@@ -0,0 +1,111 @@
...
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools]
+packages = ["splunklib", "splunklib.modularinput", "splunklib.searchcommands"]
+
+[tool.setuptools.dynamic]
+version = { attr = "splunklib.__version__" }
+
+# https://docs.astral.sh/ruff/configuration/
+# [tool.ruff]
```
mateusz834: leftover comment?

.gitignore
```diff
@@ -1,24 +1,340 @@
...
+*.sage.py
+
+# Environments
+.env
+.venv
+env/
 venv/
-.venv/
-.tox
-test-reports/
```
mateusz834, line 24: why removed? Tox still creates such files.

pyproject.toml
Ickerday: Actually intended to leave this here, but I can always add it later if something needs configuring

.gitignore
Ickerday: You're right, just removed the report generation from tox :D

.github/workflows/release.yml
```diff
@@ -1,29 +1,30 @@
 name: Release
-on:
-  release:
-    types: [published]
```
mateusz834: Why this was changed?

Ickerday: Oh crap, good catch. I split the CD away from release but didn't revert these lines. 😨

pyproject.toml
```diff
@@ -0,0 +1,83 @@
...
+DOCS = "https://docs.splunk.com/Documentation/PythonSDK/2.1.1"
+REPO = "https://github.com/splunk/splunk-sdk-python.git"
+
+[project]
+name = "splunk-sdk"
+dynamic = ["version"]
+description = "Splunk Software Development Kit for Python"
+readme = "README.md"
+requires-python = ">=3.7"
+license = { text = "Apache-2.0" }
```
mateusz834, line 15: Is this format valid? 

See: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

I think it needs to be just:


```
license = "Apache-2.0"
```

While we are here we can also add `license-files: ["LICENSE"]`

mateusz834: Discussed on slack.

mateusz834 (approved):
LGTM

PR #661, opened 2025-09-01
Add Dependabot automatic approval and merge

Ickerday:


mateusz834:
Isn't that conflicting with the reasoning of #630? Now we will automatically merge such PRs, making the situation the same as before #630, right? Since we will automatically upgrade such deps.

PR #667, opened 2025-09-08
Resolve warnings emitted by tests

mateusz834:
For the vast majority of warnings, I have simply added a skip, since we cannot remove tests for deprecated APIs.
See: https://docs.python.org/3.6/library/warnings.html#temporarily-suppressing-warnings

Additionally, to eliminate warnings caused by invalid test classes, I renamed two classes to avoid using the “TestNAME” pattern, as such names are automatically recognized as test classes.

### DIFF
--- a/tests/unit/searchcommands/test_search_command.py
+++ b/tests/unit/searchcommands/test_search_command.py
@@ -20,6 +20,7 @@ from unittest import main, TestCase
 
 import os
 import logging
+import warnings
 
 from io import TextIOWrapper
 
@@ -49,7 +50,7 @@ def build_command_input(getinfo_metadata, execute_metadata, execute_body):
 
 
 @Configuration()
-class TestCommand(SearchCommand):
+class CommandForTests(SearchCommand):
     required_option_1 = Option(require=True)
     required_option_2 = Option(require=True)
 
@@ -160,7 +161,7 @@ class TestSearchCommand(TestCase):
 
         ifile = build_command_input(getinfo_metadata, execute_metadata, execute_body)
 
-        command = TestCommand()
+        command = CommandForTests()
         result = BytesIO()
         argv = ["some-external-search-command.py"]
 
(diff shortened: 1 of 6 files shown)


cecyliaborek (approved):
👌

tests/unit/searchcommands/test_search_command.py
```diff
@@ -184,8 +185,8 @@ def test_process_scpv2(self):
         self.assertEqual(command.required_option_2, "value_2")
 
         expected = (
-            "chunked 1.0,68,0\n"
-            '{"inspector":{"messages":[["INFO","test command configuration: "]]}}'
+            "chunked 1.0,79,0\n"
+            '{"inspector":{"messages":[["INFO","commandfortests command configuration: "]]}}'
```
cecyliaborek, line 189: I know this is probably beyond the scope of this task, but maybe we could ignore the lines of the result which contain the numbers of bytes after `chunked`?  I feel like those number are a bit hard to understand when looking at the tests, but maybe there is a point to testing them?

Something like this:
```
output = result.getvalue().decode("utf-8")

payloads = "\n".join(
    line for line in output.splitlines() if not line.startswith("chunked")
)

expected_payloads = (
    '{"inspector":{"messages":[["INFO","commandfortests command configuration: "]]}}'
    '{"finished":true}test,__mv_test\r\n'
    "data,\r\n"
    "测试,\r\n"
)

self.assertEqual(expected_payloads, payloads)
```

Ickerday (approved):

Ickerday:
BTW, could we also close #638 with this one?

tests/unit/searchcommands/test_search_command.py
mateusz834: It seems like it is the only test case that we have that actually verifies that part of the protocol. I believe that it is worth to leave it there.

PR #670, opened 2025-09-09
Remove `ResultsReader`, `six`

### LINKED ISSUE #572: Splunk SDK has dependencies not satisfied by Splunk Python
The Splunk SDK currently has a requirement for `deprecation`. However, the Python version that is bundled with Splunk Enterprise does not contain  `deprecation` and does not contain `packaging`, which is a dependency of  `deprecation`.

It is common to use the Splunk SDK for developing Splunk apps (see https://dev.splunk.com/enterprise/docs/developapps/#Splunk-tools). When using version 2.x of the Splunk SDK in apps, those apps will not function correctly because of the missing dependency. It cannot be safely assumed that Splunk enterprise administrators would manually install those dependencies. I would argue that the Splunk SDK should not contain any dependencies that are not satisfied with the Python version that is bundled with Splunk.

### LINKED ISSUE #600: No deprecation library in splunk SDK
EDIT: [This docs article](https://dev.splunk.com/enterprise/docs/developapps/createapps/appanatomy/) describes how to package dependencies with an app. This would also mean any dependencies for splunklib.

The splunk sdk does not include the deprecation library, if people were to clone this repo and run:

`import splunklib.results as results`

This returns:
`ModuleNotFoundError: No module named 'deprecation'`

This is especially important for cloud customers who cannot run a pip install and thus need to add the splunklib into a custom app.

Ickerday:
I think it's high time to get rid of the old `ResultsReader`. We deprecated it in 1.6.19 (#441), we're at 2.2.0a0 now. 

Additionally, I removed the local copy of [`six`](https://pypi.org/project/six/) and instead added it as an optional dependency group under `compat`. This might get reverted - ideally we shouldn't be needing this file, but I asked around and there might be issues with removing `six` entirely because of all apps sharing one interpreter.


splunklib/results.py
```diff
@@ -32,18 +32,9 @@
...
-
-import xml.etree.ElementTree as et
-
-from collections import OrderedDict
+from io import BufferedReader
 from json import loads as json_loads
 
-__all__ = ["ResultsReader", "Message", "JSONResultsReader"]
-
-import deprecation
```
szykol, line 45: Should we consider removing the `deprecation` dependency as well? Or do we plan to stick to this method of deprecation in the future? Right now it looks like it was the only place where it was used.

szykol (approved):

TwilightMercy:
Once this will be released in v2.1.2,
There will be no longer dependency requirements of packaging + deprecation modules?

splunklib/results.py
Ickerday: Done!

Ickerday:
@TwilightMercy correct, although we still [depend on python-dotenv](https://github.com/splunk/splunk-sdk-python/pull/670/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R30).

TwilightMercy:
> @TwilightMercy correct, although we still [depend on python-dotenv](https://github.com/splunk/splunk-sdk-python/pull/670/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R30). 

Built in in Splunk modules?

Ickerday:
> Built in in Splunk modules?

Any dependencies should be packaged with your Splunk app as described [here](https://dev.splunk.com/enterprise/docs/developapps/createapps/appanatomy/). Please let us know if you can't find the info you need in the docs.

PR #671, opened 2025-09-10
SDK-21 add deps of test apps into bin/

cecyliaborek:
According to Splunk [documentation](https://dev.splunk.com/enterprise/docs/developapps/createapps/appanatomy/) and best practices, Python apps should package their dependencies in the `bin/` directory rather than `lib/`. This avoids the need to manually add `lib/` to the Python path (e.g., using `sys.path.insert`).

This PR updates our SDK test apps to follow this approach.


.gitignore
```diff
@@ -278,4 +278,6 @@ $RECYCLE.BIN/
 # End of https://www.toptal.com/developers/gitignore/api/windows,macos,linux,pycharm+all,python
 
 .vscode/
-docs/_build/
\ No newline at end of file
+docs/_build/
+
+test-reports/
```
Ickerday: Whoops, you can safely delete that folder from your local copy instead. We removed everything regarding creating these reports a bit ago. :)

Ickerday (approved):

.gitignore
cecyliaborek: oops, didn't catch that, thanks!

mateusz834 (approved):

PR #672, opened 2025-09-11
Document `key_file`, `cert_file`, `context` params.

### LINKED ISSUE #604: Question - Do we support submitting the SSL certificate when estabilishing the service to comply with requireClientCert?
**Is your feature request related to a problem? Please describe.**

When creating the service, do we support submitting the SSL certificate so we can satisfy with requireClientCert?

When enabled, the third party connecting to splunkd must present a valid certificate:

requireClientCert = <boolean>
* Whether or not an HTTPS client which connects to a splunkd
  server must possess a certificate that a certificate authority
  signed to complete the connection.
* Multiple services can use this setting, including but not
  limited to distributed search and distributed deployment clients.
  * Splunk platform indexers must use this setting to connect to
    other Splunk platform indexers.
  * Deployment clients must present certificates to deployment
    servers before they can poll the servers for new configurations
    or applications.
* A value of "true" means that a client can connect only if it
  has a certificate that was signed by a certificate
  authority that the splunkd server trusts.
* A value of "false" means that there is no certificate requirement
  to connect to services on another Splunk platform instance.
* Default: false

I have found the following issue from some years ago:

https://github.com/splunk/splunk-sdk-python/issues/123

**Describe the solution you'd like**

- Do we support it now?
- If we do not, could this be implemented in the near future?

**Describe alternatives you've considered**

**Additional context**

mateusz834:
Also add docs for all the other params, that we were missing in places like `connect()`.

Fixes #604

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -484,6 +484,12 @@ class Context:
     :type verify: ``Boolean``
     :param self_signed_certificate: Specifies if self signed certificate is used
     :type self_signed_certificate: ``Boolean``
+    :param `key_file`: Path to a PEM-encoded private key.
+    :type key_file: ``string``
+    :param `cert_file`: Path to a PEM-encoded X509 certificate chain.
+    :type cert_file: ``string``
+    :param `context`: Custom SSLContext used with the HTTPSConnection, requires verify=True.
+    :type context: ``SSLContext``
     :param sharing: The sharing mode for the namespace (the default is "user").
     :type sharing: "global", "system", "app", or "user"
     :param owner: The owner context of the namespace (optional, the default is "None").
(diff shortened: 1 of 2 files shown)


splunklib/binding.py
```diff
@@ -1505,16 +1521,16 @@ def handler(key_file=None, cert_file=None, timeout=None, verify=False, context=N
     """This class returns an instance of the default HTTP request handler using
     the values you provide.
 
-    :param `key_file`: A path to a PEM (Privacy Enhanced Mail) formatted file containing your private key (optional).
+    :param `verify`: Set to False to disable SSL verification on https connections.
```
szykol: small question: any particular reason why this description of `verify` parameter is different from others? we could be consistent and use one

szykol (approved):

splunklib/binding.py
mateusz834: Good catch, changed it to be consistent.

PR #673, opened 2025-09-11
Don't specify Host header explicitly

mateusz834:
Just let the http lib handle that for us, we shouldn't be doing that here at all.

Currently we are not handling punycode (if used) and ports, this way we don't have to care about this and just let the lib do the best thing.


mateusz834:
Verified with wireshark:

image

The Header is still present, and solves the issue in #545.

szykol (approved):

PR #681, opened 2025-11-03
Expose published and author fields

### LINKED ISSUE #619: client.Jobs.list() does not return all fields
**Describe the bug**

When I call `service.jobs.list(search="my_savedsearch")` and I look at the returned Job object, I can't find the `published` and `author` keys.

**To Reproduce**

```py
job = service.jobs.list(search="my_savedsearch")
job.state.published -> raise KeyError
job.state.author -> raise KeyError
```

**Expected behavior**

Must return the date and the user who launched the search

**Logs or Screenshots**

In comparison, I get the values by making the query manually:

```py
print(json.loads(s.get("/services/search/v2/jobs", output_mode="json", search="search=my_savedsearch").body.read().decode()))
```

Also, there's a lot more information in the `searchTelemetry` key that the Job object.

It's possible that the “state” processing doesn't take everything into account.

**Splunk (please complete the following information):**
- Version: 9.3.2
- OS: Fedora 41
- Deployment: Single instance

 - Version: 2.1.0
 - Language Runtime Version: python 3.13
 - OS: Mac 15.5

mateusz834:
Fixes #619

### DIFF
--- a/splunklib/client.py
+++ b/splunklib/client.py
@@ -279,6 +279,8 @@ def _parse_atom_entry(entry):
             "fields": metadata.fields,
             "content": content,
             "updated": entry.get("updated"),
+            "published": entry.get("published"),
+            "author": entry.get("author"),
         }
     )
 
--- a/tests/integration/test_job.py
+++ b/tests/integration/test_job.py
@@ -17,6 +17,7 @@
 from io import BytesIO
 from pathlib import Path
 from time import sleep
+from datetime import datetime
 
 import io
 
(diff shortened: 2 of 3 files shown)


szykol (approved):

splunklib/client.py
```diff
@@ -279,6 +279,8 @@ def _parse_atom_entry(entry):
             "fields": metadata.fields,
             "content": content,
             "updated": entry.get("updated"),
+            "published": entry.get("published"),
+            "author": entry.get("author", {}).get("name"),
```
Ickerday: Might be overzealous, but I was thinking of changing this to `authorName` or leaving in the `author.name` structure. Maybe we'd be able to avoid potential API surface changes this way?
Do we know if users could want something more (or is there anything more) than an author's name?

mateusz834: >  Do we know if users could want something more (or is there anything more) than an author's name?

So, I looked at the API and is seem like depending on the `output_mode` this schema looks differently.

XML:

```xml
<author>
  <name>admin</name>
</author>
```

JSON:

```json
"author": "admin"
```

Based on that, my assumption is that we will never get an additional field in that object. I think we should keep the current form in case we might want to migrate to JSON APIs in the SDK. WDYT?

mateusz834: Decided offline that we go with `author.name`.

Issue #577, opened 2024-05-27
Add ability to send custom `PUT` and `PATCH` requests with JSON body


moschi:
**Bug description**
We have a REST endpoint to update a datastructure that requires the use of the HTTP PUT method.
We're using `request` from the `Context` class to send a request to that endpoint.
On binding.py#L876, the `_encode` method is called, whose result is used to overwrite the original `body` parameter of the method. This causes the request-body to be url-encoded and url-formatted.

**Steps to reproduce the behavior**:

Send a HTTP PUT request with a JSON body to an endpoint, using the `request` method of the `Context` class.

**Expected behavior**

The `request` method correctly escapes a JSON body, or provides instructions on how to call it to prevent the URL-encoding.

**Logs or Screenshots**

Currently, our REST endpoint receives a payload similar to this:

```
event=%7B%27_key%27%3A+%2766548f65179fc9792600b193%27%2C+%27tenant_uid%27%3A+%27default%27%7D&update_objects=%7B%27attribute%27%3A+status%2C+%27value%27%3A+%2766436bc36265a3697e049fe6%27%7D&comment=Updated
```
which does not decode to a valid JSON structure:
```
event={'_key':+'66548f65179fc9792600b193',+'tenant_uid':+'default'}&update_objects={'attribute':+status,+'value':+'66436bc36265a3697e049fe6'}&comment=Updated
```

**Splunk:**
- Version: Enterprise 9.1.1
- OS: RHEL 8.8 (Ootpa)
- Deployment: single-instance

**SDK:**
 - Version: 1.7.4, bug persists in 2.0.1
 - Language Runtime Version: python 3.7
 - OS: RHEL 8.8 (Ootpa)

seiimonn:
Hi team!

Could we please get a response on this?
It seems to be a minor fix!

Thank you!

Ickerday:
Dismissed as stale. Please tell us if the issue persists!

seiimonn:
@Ickerday, this is not stale and still makes using the SDK infeasible.

Ickerday:
Thanks @seiimonn, we're taking a look!

mateusz834:
Could you provide us with a complete reproducer?

From the description i assume you are doing something like:

```py
service.request("test", method="PUT", body={"key": json.dumps({"key": 1})})
```

I observe that in such case we are sending following HTTP body:

```
key=%7B%22key%22%3A+1%7D
```

Which decodes successfully:
```py
[...]

PR #683, opened 2025-11-05
Support setting string body in the generic `request` method

mateusz834:
plus while here add test for the method-based request helpers

Updates #577

### DIFF
--- a/splunklib/binding.py
+++ b/splunklib/binding.py
@@ -938,7 +938,11 @@ class Context:
             str(mask_sensitive_data(dict(all_headers))),
             mask_sensitive_data(body),
         )
-        if body:
+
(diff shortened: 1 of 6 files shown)


splunklib/binding.py
```diff
@@ -938,7 +938,12 @@ def request(
             str(mask_sensitive_data(dict(all_headers))),
             mask_sensitive_data(body),
         )
-        if body:
+
+        if isinstance(body, str):
+            if method.upper() == "GET":
+                raise Exception("unable to set string body on GET method request")
```
Ickerday:
```suggestion
            assert method.upper() == "GET", "Unable to set body on GET request"
```

Ickerday (approved):

PR #694, opened 2026-01-19
Upstream changes from the internal mirror

Ickerday:
- Update the License field from a deprecated format to a simple string
- Simplify the dependency groups
- Bump up the minimum Python version to 3.9
- Bump up 3rd party package versions
- Update files with internal mirror versions


pyproject.toml
```diff
@@ -11,14 +11,12 @@ name = "splunk-sdk"
...
-license = { text = "Apache-2.0" }
+requires-python = ">=3.13.0, <3.14"
+license = "Apache-2.0"
 authors = [{ name = "Splunk, Inc.", email = "devinfo@splunk.com" }]
 keywords = ["splunk", "sdk"]
 classifiers = [
   "Programming Language :: Python :: 3",
-  "Programming Language :: Python :: 3 :: Only",
-  "Programming Language :: Python :: 3.7",
   "Programming Language :: Python :: 3.9",
```
szykol, line 22: I guess this line is to be deleted too

pyproject.toml
```diff
@@ -11,14 +11,12 @@ name = "splunk-sdk"
 dynamic = ["version"]
 description = "Splunk Software Development Kit for Python"
 readme = "README.md"
-requires-python = ">=3.7"
-license = { text = "Apache-2.0" }
+requires-python = ">=3.13.0, <3.14"
```
mateusz834:
```suggestion
requires-python = ">=3.13.0"
```

Why `<3.14?`

Ickerday: That was intentional - I've had some packages scream incompatibility with 3.14 but I can't reproduce that now, so that might've been fixed in the meantime.

Makefile
```diff
@@ -35,14 +63,22 @@ docker-start: docker-up docker-ensure-up
...
+	docker-compose rm -f -s
 
 .PHONY: docker-refresh
-docker-refresh: docker-remove docker-start
\ No newline at end of file
+docker-refresh: docker-remove docker-start
+
+.PHONY: docker-splunk-restart
+docker-splunk-restart:
+	docker exec -it splunk-instance sh -c '/opt/splunk/bin/splunk restart'
```
mateusz834:
```suggestion
	docker exec -it splunk sh -c '/opt/splunk/bin/splunk restart'
```

Or has something changed that causes that?

Same below.

Makefile
```diff
@@ -35,14 +63,22 @@ docker-start: docker-up docker-ensure-up
...
+
+.PHONY: docker-tail-python-log
+docker-tail-python-log:
+	docker exec splunk-instance tail /opt/splunk/var/log/splunk/python.log
```
mateusz834:
```suggestion
	docker exec splunk sudo tail /opt/splunk/var/log/splunk/python.log
```

Makefile
Ickerday: Good catch! I took it from the PoC repo which had a bit different setup and didn't confirm we ever ported it back to this repo. :D

szykol (approved):

PR #695, opened 2026-01-19
Deprecate the `encr_password` field from `storage/passwords`

Ickerday:
Splunk is deprecating the `encr_password` field from `/services/storage/passwords`. We need to mark the field as deprecated to instruct users to move away from it. 

More info: https://help.splunk.com/en/splunk-enterprise/leverage-rest-apis/rest-api-reference/10.2/access-endpoints/access-endpoint-descriptions#storage.2Fpasswords

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,7 @@ services:
     splunk:
         image: "splunk/splunk:${SPLUNK_VERSION}"
         container_name: splunk
+        platform: linux/amd64
         environment:
             - SPLUNK_START_ARGS=--accept-license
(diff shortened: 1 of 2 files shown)


szykol (approved):

.github/workflows/test.yml
```diff
@@ -11,12 +11,6 @@ jobs:
         python-version: [3.9]
         splunk-version: [9.4, latest]
         include:
```
mateusz834, line 13: .

mateusz834 (commented):
Thanks!

mateusz834 (approved):

PR #699, opened 2026-02-09
Adjust top-level comments in all files

Ickerday:
- Adjusted copyright dates
- Removed obsolete [# coding=utf8](https://peps.python.org/pep-0263/) statements (it's been a long time since Python started treating all files as UTF-8)
- Removed shebangs


mateusz834 (approved):
Thanks

PR #724, opened 2026-04-16
Make id required (non-nullable) for tool/subagent/output calls

mateusz834:
Also while here, move the id field to be first.

### DIFF
--- a/splunklib/ai/engines/langchain.py
+++ b/splunklib/ai/engines/langchain.py
@@ -481,10 +481,45 @@ class LangChainAgentImpl(AgentImpl[OutputT]):
 
                 return call
 
+        class _CheckCallIDMiddleware(LC_AgentMiddleware):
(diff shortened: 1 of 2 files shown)


Ickerday (approved):

splunklib/ai/engines/langchain.py
```diff
@@ -481,10 +481,45 @@ def unpack_tool_call(self, call: LC_ToolCall) -> LC_ToolCall:
 
                 return call
 
+        class _CheckCallIDMiddleware(LC_AgentMiddleware):
+            def _check_has_call_id(self, msg: LC_AIMessage) -> None:
+                for call in msg.tool_calls:
+                    if call["id"] is None:
```
mateusz834:
```suggestion
                    if not call["id"]:
```

szykol (approved):

PR #727, opened 2026-04-17
Add GoogleModel support

szykol:
This PR adds support to Google Models via:
- Gemini API
- VertexAI

The actual backend used depends on the parameters passed to GoogleModel

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -72,16 +72,6 @@
                 }
             }
         ],
-        "./splunklib/ai/model.py": [
-            {
-                "code": "reportDeprecated",
-                "range": {
-                    "startColumn": 24,
-                    "endColumn": 31,
-                    "lineCount": 1
-                }
-            }
-        ],
         "./splunklib/ai/serialized_service.py": [
             {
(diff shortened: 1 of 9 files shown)


Ickerday (approved):

szykol:
update:

Those changes were tested with Gemini models through Gemini API.
The thought signature issue that we had when using Gemini models through OpenAI compatible API is (partially) gone now.
However, our SDK does not preserve the though signatures in it's own types and this works only because the `langchain-google-genai` package uses dummy thought signatures when original ones (generated by the model) are missing.

I imagine we're gonna need to include the original though signatures when passing new messages to the Gemini model, but this requires a couple of changes to the SDK code - mainly how we handle content of `AIMessage`. Currently we're dumping everything to string, but some model providers return content blocks instead of strings there. Gemini models include thought signatures inside those content blocks (if not calling tools). If calling tools, the though signature is stored in the `additional_kwargs` field of `AIMessage`.

Once we'll fix handling content in `AIMessage`s we can continue with this PR. It's already in progress.

mateusz834 (approved):

PR #740, opened 2026-04-23
Snapshot record then reply testing

mateusz834:
This change implements snapshot, record then reply testing for all integration tests.

Re-enables recently skipped (flaky) tests:

- `test_agent_understands_other_agents` (snapshot was edited manually)
- `test_supervisor_resumes_subagent_thread_across_invocations`
- `test_supervisor_resumes_subagent_thread_across_invocations_structured`

Introduces a deterministic thread_id mock generator, such that snapshots are deterministic, for:

- `test_supervisor_resumes_subagent_thread_across_invocations`
- `test_supervisor_resumes_subagent_thread_across_invocations_structured`

Modified `test_tool_execution_service_access` using tool middleware, to make the test deterministic.

E2E tests still call the LLMs directly.

### DIFF
--- a/tests/ai_testlib.py
+++ b/tests/ai_testlib.py
@@ -1,7 +1,21 @@
-from typing import override
+import functools
+import inspect
+import json
+import os
+from collections.abc import Callable, Coroutine
+from typing import Any, override
+from unittest.mock import patch
+from urllib import parse
+
+import vcr
+from vcr.config import RecordMode
+from vcr.request import Request
+
 from splunklib.ai.model import PredefinedModel
 from tests.ai_test_model import InternalAIModel, TestLLMSettings, create_model
 from tests.testlib import SDKTestCase
+
+REDACTED_APP_KEY = "[[[--APPKEY-REDACTED-]]]"
 
 
 class AITestCase(SDKTestCase):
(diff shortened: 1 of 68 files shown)


tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
...
+                    assert settings.internal_ai.token_url.lower() not in out.lower()
+                    assert settings.internal_ai.client_id.lower() not in out.lower()
+                    assert settings.internal_ai.client_secret.lower() not in out.lower()
+
+                    return out
+
+            def _before_record_request(request: Request) -> Request | None:
+                url = parse.urlparse(request.uri)  # pyright: ignore[reportUnknownArgumentType, reportUnknownVariableType]
+                if url.hostname == internal_ai_hostname:
+                    request.headers = {}
```
splunk-dtaborski, line 135: Shouldn't we keep some specific headers for like Authorization and keep comparing them?

tests/integration/ai/test_agent.py
```diff
@@ -40,14 +40,15 @@
...
     subagent_middleware,
 )
-from tests.ai_testlib import AITestCase
+from tests.ai_testlib import AITestCase, ai_snapshot_test
 
 OPENAI_BASE_URL = "http://localhost:11434/v1"
 OPENAI_API_KEY = "ollama"
 
 
 class TestAgent(AITestCase):
```
splunk-dtaborski, line 49: Not sure if correct assumption but i think most of the time will be decorating all tests in the file. Maybe we could add one more decoration on class level that would add function decorator to each test automatically.

High-level idea:
instead of
```
class TestIntegration:
  @ai_snapshot_test()
  def func1():
    ...
  @ai_snapshot_test()
  def func2():
    ...
```

Do
```
@snapshot_test()
class TestIntegration:
    def func1():
      ...

    def func2():
      ...
```

I dont say it has to be specifically in this PR however i wanted to point it out.

tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
         model = await create_model(self.test_llm_settings)
         self._model = model
         return model
+
+
+def ai_snapshot_test() -> Callable[
```
splunk-dtaborski, line 61: [nitpick] the name suggests it can be used only for AI. What do you think about record_snapshot name?

splunk-dtaborski (commented):
I really like the solution.

tests/ai_testlib.py
mateusz834: > the name suggests it can be used only for AI.

We do filtering here, so i think it only works for AI 😄 .

tests/integration/ai/test_agent.py
mateusz834: That is a nice idea, but i feel that the complexity of it might be huge, i.e. what if we want to have some test that is not snapshoted, then we need some kind of exclude decorator.
I also see us running these tests with multiple snapshots, say from openai, gemini, anthropic (as subtests, created by this decorator) and there might be a need to say skip specific providers from specific tests or such. For example in test_structured_output.pl for models that do not support provider strategy.

CC @szykol @Ickerday for opinions.

szykol: I think if we wanted to achieve something like this we could just add a fixture instead.
The fixtures can be autoused, so that all files in the scope of the fixture behave the same.

But I agree with Mateusz I don't see an issue with having those tests explicitly say they are snapshot. Also turning of the snapshotting of a specific test is a matter of removing the decorator, no further work needed.

tests/integration/ai/test_agent.py
```diff
@@ -193,7 +200,7 @@ class NicknameGeneratorInput(BaseModel):
             result = await supervisor.invoke(
                 [
                     HumanMessage(
-                        content="hi, my name is Chris. Generate a nickname for me",
+                        content="i, my name is Chris. Generate a nickname for me",
```
szykol: looks like a typo by mistake

tests/integration/ai/test_agent_mcp_tools.py
```diff
@@ -99,15 +104,41 @@ async def test_tool_execution_structured_output(self) -> None:
         os.path.join(os.path.dirname(__file__), "testdata", "tool_context.py"),
     )
     @patch("splunklib.ai.agent._testing_app_id", "app_id")
+    @ai_snapshot_test()
```
szykol, line 107: is calling the decorator needed?

`@ai_snapshot_test` looks a little easier to the eye

szykol: but this requires changing the code in the decorator

tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
...
+        async def wrapper(self: AITestCase, *args: Any, **kwargs: Any) -> None:
+            settings = self.test_llm_settings
+            assert settings.internal_ai is not None
+
+            internal_ai_hostname = parse.urlparse(
+                settings.internal_ai.base_url
+            ).hostname
+            assert internal_ai_hostname is not None
+
+            REDACTED_APP_KEY = "[[[--APPKEY-REDACTED-]]]"
```
szykol: nit: this could be moved to the top of file

tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
...
+
+                    # Assert that nothing is leaking into the public snapshots.
+                    assert internal_ai_hostname not in out.lower()
+                    assert settings.internal_ai.app_key.lower() not in out.lower()
+                    assert settings.internal_ai.base_url.lower() not in out.lower()
+                    assert settings.internal_ai.token_url.lower() not in out.lower()
+                    assert settings.internal_ai.client_id.lower() not in out.lower()
+                    assert settings.internal_ai.client_secret.lower() not in out.lower()
+
+                    return out
```
szykol, line 130: curious if we should look for the other secrets for our testing LLM before serializing - to make sure we do not leak anything by mistake when creating new snapshots.

or add some ci stage for that.

WDYT?

tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
...
+                    "host",
+                    "port",
+                    "path",
+                    "query",
+                    "jsonbody",
+                ],
+                before_record_request=_before_record_request,
+                before_record_response=_before_record_response,
+                # record_on_exception=False,
+                # drop_unused_requests=True,
```
szykol: to be removed if not needed

tests/ai_testlib.py
```diff
@@ -42,3 +54,149 @@ async def model(self) -> PredefinedModel:
...
+
+            with my_vcr.use_cassette(snapshot_filename):  # pyright: ignore[reportGeneralTypeIssues]
+                await fn(self, *args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
+
+def deterministic_thread_ids() -> Callable[
```
szykol, line 178: that's cool, we don't need to change the test body to accommodate that.

tests/ai_testlib.py
mateusz834: I want to keep them since I have found them as useful options to toggle when you are adding say a new test or just experimenting with this recordings, They are not used now but i uncommented them multiple times, so might be worth, WDYT?

tests/ai_testlib.py
mateusz834: > curious if we should look for the other secrets for our testing LLM before serializing - to make sure we do not leak anything by mistake when creating new snapshots.

Do you have anything other in mind to check? We are only recording LLM http requests, and i think we are checking everything.

tests/integration/ai/test_agent_mcp_tools.py
mateusz834: I made it a function just so we can accept parameters, I believe we might want to have them once we add more LLMs to our integration tests.

For example if we would want to say skip recording for LLM X of snapshots, since we don't have recording for some LLM on some specific test.

tests/ai_testlib.py
mateusz834: I thought that it makes sense, but now i realized that that would only test our test suite since the header that is used for our internal ai does not match how our end used will use this, say with OpenAI API directly.

 https://github.com/splunk/splunk-sdk-python/blob/b1219b2507f8a6d3b5f33f60a294e302531bc897/tests/ai_test_model.py#L49-L51
 
I don't believe now that the complexity of adding this is worth.

szykol: I think we should have a small suite that always runs on real LLMs that is separate from those snapshot tests. Those would run to check if we have some regressions etc. And those would make sure we're passing proper auth headers etc.

mateusz834: > And those would make sure we're passing proper auth headers etc.

For now these are E2E tests.

tests/integration/ai/test_agent_mcp_tools.py
szykol: sure thing, also making the decorator working both ways `@decorator` or `@decorator()` was not the greatest experience for me last time I tried this so if we plan to have some parameters there in the future i'm okay with this.

tests/ai_testlib.py
szykol: let's say we add some new parameters to our internal ai model, or the response changes and we re-record, without some automatic checks we rely on ourselves making sure we do not leak that info by mistake because it was overlooked.

but on the other it's hard to precisely say what could go wrong so maybe let's stick to double checking while adding new recordings

mateusz834: > but on the other it's hard to precisely say what could go wrong so maybe let's stick to double checking while adding new recordings

My main point here was to automate checking all the added recordings that are part of this PR. I don't expect new PR to contain that huge of amount of snapshots thus will be easier to review them individually, as you mentioned.

tests/ai_testlib.py
szykol: I don't know, those parameters make sense to you since you've worked with vcr, but to me, at least at first sight, it looks like code that was left by mistake and should be removed.

I agree it's easy to uncomment and use them, but without context they do not add much value to me.

This is testing code so I'm not too picky about this, maybe add some comment above them and explain what they do or say "uncomment if you want to...".

Up to you tho!

tests/ai_testlib.py
szykol: sure, my main point was to brainstorm how to make sure we do not leak anything, but I guess we have to rely on ourselves here.

tests/ai_testlib.py
mateusz834: Ok, decided to uncomment them, sounds like a better default.

szykol (approved):

PR #741, opened 2026-04-23
Improve handling `AIMessage.content`

szykol:
The content of `AIMessage` could differ between different providers.

For example:
* OpenAI gpt-5-nano responds with single string
* Gemini gemini-3-* models respond with list of `content-block`s. Those are dictionaries that include additional information.

An example of Gemini's content-block could look like this:

```
{
    "type": "text",
    "text": "test-content-block",
    "extras": {
          # simulate gemini model returning thought signature in extra field of text content block
          "signature": "EjQKMgEMOdbHDmsQ+BTM6duYJ43i5npxkpn28Ir0VjD1p6w4fUqIdYszIcWx+XcqAW1a8E+Q"
      }
}
```

This content-block contains additional information such as the thought signatures that the Gemini models use.
Additionaly, the Gemini models include the thought signatures in langchain's `AIMessage.additional_kwargs` when issuing a tool call - this is also addressed here by adding an `extras` field to the SDK `AIMessage`.

All those changes should accomodate to the model providers sending additional information along their responses.

> **NOTE**: This change slightly changes the API, since we change the type of `AIMessage.content` from `str` to `str | list[str | ContentBlock]`.
> It also adds new ContentBlock type to note that the response from LLM is structured (not a string).

Currently, only the `TextBlock` is supported, since I didn't see any other one returned from the LLMs I had access to. Other "unsupported" content blocks are stored in `OpaqueBlock`s which are created to persist blocks that are returned by LLM, but not useful for the SDKs. Those would be returned back to the LLM so that the context is preserved. We can add support to other blocks if needed.

The developers should handle possible formats of responses by checking the type or use structured outputs to make sure every model responds with the same format.

### DIFF
--- a/splunklib/ai/engines/langchain.py
+++ b/splunklib/ai/engines/langchain.py
@@ -77,7 +77,9 @@ from splunklib.ai.messages import (
     AgentResponse,
     AIMessage,
     BaseMessage,
+    ContentBlock,
     HumanMessage,
+    OpaqueBlock,
     OutputT,
     StructuredOutputCall,
     StructuredOutputMessage,
@@ -87,6 +89,7 @@ from splunklib.ai.messages import (
     SubagentStructuredResult,
     SubagentTextResult,
     SystemMessage,
+    TextBlock,
     ToolCall,
     ToolFailureResult,
     ToolMessage,
@@ -955,7 +958,7 @@ class _Middleware(LC_AgentMiddleware):
         return LC_ToolMessage(
             name=_normalize_agent_name(call.name),
             tool_call_id=call.id,
-            content=content,
+            content=_map_content_to_langchain(content),
             status=status,
             artifact=sdk_result,
         )
@@ -1089,7 +1092,10 @@ def _convert_model_response_to_model_result(
     # This invariant is asserted via ModelResponse.__post_init__
     assert len(resp.message.structured_output_calls) <= 1
 
-    lc_message = LC_AIMessage(content=resp.message.content)
+    lc_message = LC_AIMessage(
+        content=_map_content_to_langchain(resp.message.content),
+        additional_kwargs=resp.message.extras or {},
+    )
     # This field can't be set via __init__()
     lc_message.tool_calls = [_map_tool_call_to_langchain(c) for c in resp.message.calls]
 
@@ -1164,7 +1170,7 @@ def _convert_tool_message_to_lc(
         name=name,
         tool_call_id=message.call_id,
         status=status,
-        content=content,
+        content=_map_content_to_langchain(content),
         artifact=artifact,
     )
 
@@ -1247,9 +1253,10 @@ def _convert_model_result_from_lc(model_response: LC_ModelCallResult) -> ModelRe
         ai_message = model_response
         structured_response = None
 
+    additional_kwargs = cast(dict[str, Any], ai_message.additional_kwargs)
     return ModelResponse(
         message=AIMessage(
-            content=ai_message.content.__str__(),
+            content=_map_content_from_langchain(ai_message.content),  # pyright: ignore[reportUnknownArgumentType]
             calls=[
                 _map_tool_call_from_langchain(tc)
                 for tc in ai_message.tool_calls
@@ -1264,6 +1271,7 @@ def _convert_model_result_from_lc(model_response: LC_ModelCallResult) -> ModelRe
                 for tc in ai_message.tool_calls
                 if tc["name"].startswith(TOOL_STRATEGY_TOOL_PREFIX)
             ],
+            extras=additional_kwargs,
         ),
         structured_output=structured_response,
     )
@@ -1426,6 +1434,28 @@ def _is_agent_name_valid(name: str) -> bool:
     return set(name).issubset(AGENT_NAME_ALLOWED_CHARS)
 
 
+def _parse_content_block(block: str | ContentBlock) -> str | None:
+    match block:
+        case TextBlock():
+            return block.text
+        case str():
+            return block
+        case _:
+            return None
+
+
+def _parse_content(content: str | list[str | ContentBlock]) -> str:
+    """Parses the content from AIMessage and builds a single string our of it"""
+    if isinstance(content, str):
+        return content
+
+    return " ".join(
+        parsed_block
+        for block in content
+        if (parsed_block := _parse_content_block(block))
+    )
+
+
 def _agent_as_tool(agent: BaseAgent[OutputT]) -> StructuredTool:
     if not agent.name:
         raise AssertionError("Agent must have a name to be used by other Agents")
@@ -1437,7 +1467,10 @@ def _agent_as_tool(agent: BaseAgent[OutputT]) -> StructuredTool:
 
     async def invoke_agent(
         message: HumanMessage, thread_id: str | None
-    ) -> tuple[OutputT | str, SubagentStructuredResult | SubagentTextResult]:
+    ) -> tuple[
+        OutputT | str,
+        SubagentStructuredResult | SubagentTextResult,
+    ]:
         result = await agent.invoke([message], thread_id=thread_id)
 
         if agent.output_schema:
@@ -1446,9 +1479,8 @@ def _agent_as_tool(agent: BaseAgent[OutputT]) -> StructuredTool:
                 structured_output=result.structured_output.model_dump(),
             )
 
-        return result.final_message.content, SubagentTextResult(
-            content=result.final_message.content
-        )
+        text_content = _parse_content(result.final_message.content)
+        return text_content, SubagentTextResult(content=text_content)
 
     InputSchema = agent.input_schema
     if InputSchema is None:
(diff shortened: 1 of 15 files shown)


splunklib/ai/engines/langchain.py
```diff
@@ -1433,7 +1441,10 @@ def _agent_as_tool(agent: BaseAgent[OutputT]) -> StructuredTool:
 
     async def invoke_agent(
         message: HumanMessage, thread_id: str | None
-    ) -> tuple[OutputT | str, SubagentStructuredResult | SubagentTextResult]:
+    ) -> tuple[
+        OutputT | str | list[str | ContentBlock],
```
mateusz834: I am uncertain here, shouldn't we extract all the text contents and only return these?

splunklib/ai/engines/langchain.py
```diff
@@ -1564,11 +1590,66 @@ def _map_tool_call_to_langchain(call: ToolCall | SubagentCall) -> LC_ToolCall:
...
+) -> str | ContentBlock:
+    if isinstance(block, str):
+        return block
+
+    match block.get("type"):
+        case "text":
+            return TextBlock(
+                text=block["text"],
+                extras=block.get("extras"),
+            )
```
mateusz834, line 1638: What if providers add different fields here, we will ignore them?

splunklib/ai/messages.py
```diff
@@ -85,12 +110,15 @@ class AIMessage(BaseMessage):
     """
 
     role: Literal["assistant"] = field(default="assistant", init=False)
-    content: str
+    content: str | list[str | ContentBlock]
```
mateusz834, line 129: What if we made this a:

```py
    content: str | list[str | dict]
```

or

```py
    content: str | list[str | OpaqueBlock]
```

just as LC does? and provide a `contect_block` property that converts this to blocks (same just as LC)?

```
class ContentBlock:
    extras: dict[str, Any] | None
    pass

class TextBlock(ContentBlock):
     text: str
```

or instead of `contect_block` we could provide `texts` property that returns a `list[str]`, extracted from the content? And then later if needed we can figure out the blocks thing.

splunklib/ai/engines/langchain.py
szykol: sounds like this is not provider specific but LC specific, so all of the "extra" properties should be stored here. The `extras` field is just a dict so we can expand it later if needed, and it's more of an internal field (it's used by the langchain layer) so I don't see an issue of changing it in the future.

splunklib/ai/messages.py
szykol: I imagined the solution to be more explicit, so that the SDK controls the type of content blocks that are "supported", leaving the OpaqueBlock strictly for preserving the data between the SDK and langchain layer, so that the LLMs receive all the context in the next call.

This makes them explicitly ignore the OpaqueBlock and if any new model returns some new content block, this would fall into the OpaqueBlock, meaning the already existing code shouldn't break once this happens.

Of course someone could use the OpaqueBlock if they wanted to and create some logic around it's contents, but they **should** expect this code to be changed in case SDK adds first class support to this specific block. And I don't see the issue with breaking such code in the future, since it wasn't officially supported.

I don't think we can future proof everything, at least without creating some other potential issues.

By making the `OpaqueBlock` or in this case more of a base `ContentBlock` we could add a helper method for getting the text list from them (if possible).
I'm worried this could introduce issues in the developers code as the content-blocks and their contents might change in the future, but I'm open to discussion.
However, In this case we could also just have the `TextBlock` accept any content-block that contains the `text` field, or any other content-block that you can get the text from. And leave the OpaqueBlock as is.

splunklib/ai/engines/langchain.py
szykol: we just forward the return of inner `invoke_agent`, what's wrong here?

mateusz834: My point is that we should only forward the text response as an output of an subagent, right?

mateusz834: If there are others i don't think we should forward them??

szykol: would you want to filter them out or assert if any other block is OpaqueBlock?

szykol: switched it back so that the subagent text result returns only strings: https://github.com/splunk/splunk-sdk-python/compare/ad69e2ca3a395d629d5d796558031de340c42f31..1a7d1be1893f2a1d8600c37c9b12a7bc8315cf4b

szykol: and https://github.com/splunk/splunk-sdk-python/compare/1a7d1be1893f2a1d8600c37c9b12a7bc8315cf4b..c383f8c20df7d03816171235ec408ecc4e4b6239

splunklib/ai/engines/langchain.py
```diff
@@ -1442,8 +1475,9 @@ async def invoke_agent(
                 structured_output=result.structured_output.model_dump(),
             )
 
-        return result.final_message.content, SubagentTextResult(
-            content=result.final_message.content
+        text_content = _parse_content(result.final_message.content)
+        return text_content, SubagentTextResult(
+            content=_parse_content(result.final_message.content)
```
mateusz834:
```suggestion
            content=text_content
```

No need to call this twice for same arg.

splunklib/ai/engines/langchain.py
```diff
@@ -1452,13 +1486,19 @@ async def invoke_agent(
 
             async def _run(  # pyright: ignore[reportRedeclaration]
                 content: str, thread_id: str
-            ) -> tuple[OutputT | str, SubagentStructuredResult | SubagentTextResult]:
+            ) -> tuple[
+                OutputT | str | list[str | ContentBlock],
```
mateusz834:
```suggestion
                OutputT | str
```

splunklib/ai/engines/langchain.py
```diff
@@ -1452,13 +1486,19 @@ async def invoke_agent(
...
+                SubagentStructuredResult | SubagentTextResult,
+            ]:
                 return await invoke_agent(HumanMessage(content=content), thread_id)
         else:
 
             async def _run(  # pyright: ignore[reportRedeclaration]
                 content: str,
-            ) -> tuple[OutputT | str, SubagentStructuredResult | SubagentTextResult]:
+            ) -> tuple[
+                OutputT | str | list[str | ContentBlock],
```
mateusz834:
```suggestion
                OutputT | str
```

splunklib/ai/engines/langchain.py
```diff
@@ -1484,15 +1527,17 @@ async def invoke_agent_structured(
...
+        text_content = _parse_content(result.final_message.content)
+        return text_content, SubagentTextResult(content=text_content)
 
     if agent.conversation_store:
 
         async def _run(
             **kwargs: Any,  # noqa: ANN401
-        ) -> tuple[OutputT | str, SubagentStructuredResult | SubagentTextResult]:
+        ) -> tuple[
+            OutputT | str | list[str | ContentBlock],
```
mateusz834:
```suggestion
            OutputT | str
```

splunklib/ai/engines/langchain.py
```diff
@@ -1512,7 +1557,10 @@ async def _run(
...
```
mateusz834:
```suggestion
            OutputT | str
```

splunklib/ai/messages.py
```diff
@@ -85,12 +110,15 @@ class AIMessage(BaseMessage):
...
 
     calls: Sequence[ToolCall | SubagentCall]
     structured_output_calls: Sequence[StructuredOutputCall] = field(
         default_factory=tuple
     )
+    # Backend-specific metadata (e.g. provider additional_kwargs) not
+    # representable in the standard fields. Opaque to callers.
+    extras: dict[str, Any] | None = field(default=None)
```
mateusz834, line 135: Do we want to add some kind of note that there are no guarantees here? i.e. that this field is expected to be replied back but should not be read/modified as the contents of this might change or such?

splunklib/ai/engines/langchain.py
szykol: yeah, I fixed this in other place but not here, thanks for catching this.

splunklib/ai/engines/langchain.py
szykol: sounds like we need to do some refactoring because changing this code is painful 😓

splunklib/ai/engines/langchain.py
szykol: done

splunklib/ai/engines/langchain.py
szykol: done

splunklib/ai/engines/langchain.py
szykol: done

splunklib/ai/engines/langchain.py
szykol: done

splunklib/ai/messages.py
szykol: not sure what else to add here, it already says that it's backend-specific metadata and it's in a frozen dataclass. Happy to add if you have anything else in mind!

mateusz834: Something like:

```suggestion
    # This field contains LLM-specific metadata.
    # It should be returned to the LLM unchanged in
    # subsequent LLM calls.
    # The contents of this field is not guaranteed to
    # be stable and might change as SDK evolves.
    extras: dict[str, Any] | None = field(default=None)
```

mateusz834 (approved):

PR #756, opened 2026-04-30
Add `StructuredOutputRetryLimitMiddleware` and default retry limit

mateusz834:
Regenerated `test_agent_understands_other_agents.json`, since the previous recoding hit that default limit.

### DIFF
--- a/splunklib/ai/base_agent.py
+++ b/splunklib/ai/base_agent.py
@@ -23,9 +23,11 @@ from pydantic import BaseModel
 from splunklib.ai.conversation_store import ConversationStore
 from splunklib.ai.hooks import (
     DEFAULT_STEP_LIMIT,
+    DEFAULT_STRUCTURED_OUTPUT_RETRY_LIMIT,
     DEFAULT_TIMEOUT_SECONDS,
     DEFAULT_TOKEN_LIMIT,
     StepLimitMiddleware,
+    StructuredOutputRetryLimitMiddleware,
     TimeoutLimitMiddleware,
     TokenLimitMiddleware,
 )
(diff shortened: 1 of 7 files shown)


splunklib/ai/base_agent.py
```diff
@@ -79,16 +81,24 @@ def __init__(
         self._output_schema = output_schema
         user_middleware = tuple(middleware) if middleware else ()
         user_middleware_types = {type(m) for m in user_middleware}
+
         # NOTE: we're creating separate instances per agent - TimeoutLimitMiddleware is stateful
         # and sharing one would cause agents to overwrite each other's deadline.
-        predefined: list[AgentMiddleware] = [
+        predefined_before: list[AgentMiddleware] = [
```
splunk-dtaborski, line 87: why did you dive it?

mateusz834: The `StructuredOutputRetryLimitMiddleware` must be first in the list (L97 below) since it has to execute after all other `user_middleware` have been executed, otherwise, if some `user_middleware` raised the `StructuredOutputGenerationException` we would not have catched that and the agent loop would have re-tried the structured output generation, without any re-try limits.

tests/integration/ai/test_structured_output.py
```diff
@@ -930,5 +934,148 @@ async def _model_middleware(
             assert len(result.messages) == 3
             assert result.structured_output.name == "MIKE"
 
+    @pytest.mark.asyncio
+    @ai_snapshot_test()
+    async def test_default_retry_limit(self) -> None:
+        pytest.importorskip("langchain_openai")
```
splunk-dtaborski, line 944: [nitpick] can't we add module level guard instead of repeating ? I mean this whole file depends on langchain_openai

mateusz834: We have this skip basically in every test now, if we are going to change anything related to this i think we should just remove them.

I’m not aware of any way to apply this globally, especially in a way that plays nicely with pytest, since these skips are actually executed per test and are visible in the pytest output.

splunk-dtaborski: ok

tests/integration/ai/snapshots/test_agent/TestAgent.test_agent_understands_other_agents.json
szykol, line 1: shouldn't we have additional files here, because you've added a couple of integration/snapshot tests?

szykol: discussed with @mateusz834, all the tests mock out the llm, that's why

mateusz834: No, since these LLM calls are mocked in middlewares.

szykol (approved):

PR #759, opened 2026-05-04
Move limit middlewares from `splunklib.ai.hooks` to `splunklib.ai.limits`

mateusz834:
Also changed the TokenLimitExceededException to accept an int, instead of a float.


splunklib/ai/limits.py
```diff
@@ -0,0 +1,170 @@
+from time import monotonic
```
szykol, line 15: missing copyrights section

szykol (approved):

splunklib/ai/limits.py
mateusz834: added

PR #764, opened 2026-05-05
test: Pass fake api_key to the internal ai model

szykol:
The new version of openai library (v2.34.0) requires the `api_key` to be non-empty.
This fails our e2e tests, since we didn't use the `api_key`.
This PR fixes this by passing some fake `api_key` to the `OpenAIModel` for tests.

### DIFF
--- a/tests/ai_test_model.py
+++ b/tests/ai_test_model.py
@@ -1,8 +1,4 @@
-import collections.abc
-from typing import override
-
 import httpx
-from httpx import Auth, Request, Response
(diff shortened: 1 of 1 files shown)


Ickerday (approved):

mateusz834 (approved):
Please update the description and title, since this is no longer a test.

szykol:
Sorry for poor wording, I wanted to note that this is related to tests only, not that this was used for testing.

PR #767, opened 2026-05-07
CD adjustments again

Ickerday:


.github/workflows/cd.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Generate API reference
+        run: make -C ./docs html
+      - name: Upload docs artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
+        with:
+          name: python-sdk-docs
+          path: docs/_build/html
+
+  publish-pre-release:
+    if: startsWith(github.ref, 'refs/tags/') != true
```
mateusz834:
```suggestion
    if: startsWith(github.ref, 'refs/tags/') == false
```

.github/workflows/cd.yml
```diff
@@ -0,0 +1,90 @@
...
+        with:
+          name: splunk-sdk-${{ needs.build-distributables.outputs.version }}
+          path: dist/
+      - name: Publish packages to Test PyPI
+        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+  publish-release:
+    if: startsWith(github.ref, 'refs/tags/')
```
mateusz834:
```suggestion
    if: startsWith(github.ref, 'refs/tags/') == true
```

Since above we use this form and this makes it easier to see the difference.

mateusz834 (approved):
LGTM

PR #769, opened 2026-05-08
Make limits as settings instead of implicit middlewares

mateusz834:


splunklib/ai/README.md
```diff
@@ -977,103 +977,87 @@ model = OpenAIModel(...)
...
-The hooks can stop the Agentic Loop under custom conditions by raising exceptions.
-The logic of the hook can be more advanced and include multiple conditions, for example, based on both token usage and execution time:
+The hooks can stop the Agentic Loop under custom conditions by raising exceptions, for example:
 
 ```py
 from splunklib.ai.hooks import before_model
 from splunklib.ai.middleware import AgentMiddleware, ModelRequest
 
-def token_and_step_limit(token_limit: float, step_limit: int) -> AgentMiddleware:
+def token_and_message_limit(message_limit: int) -> AgentMiddleware:
```
szykol: we no longer check for the token count in this middleware so

```suggestion
def message_limit(message_limit: int) -> AgentMiddleware:
```

splunklib/ai/agent.py
```diff
@@ -149,6 +156,7 @@ def __init__(
         output_schema: type[OutputT] | None = None,
         input_schema: type[BaseModel] | None = None,  # Only used by Subagents
         middleware: Sequence[AgentMiddleware] | None = None,
+        limits: AgentLimits = DEFAULT_AGENT_LIMITS,
```
szykol, line 159: this caught my attention because its not always a good idea to accepts objects as default parameter values. However `AgentLimits` is frozen, which makes it immutable as long as we have only primitive values in there.

mateusz834: FYI we are doing this already for `tool_settings: ToolSettings = DEFAULT_TOOL_SETTINGS`

splunklib/ai/engines/langchain.py
```diff
@@ -250,13 +258,26 @@ def __init__(self, agent: BaseAgent[OutputT]) -> None:
 
         before_user_middlewares, after_user_middlewares = _debugging_middleware(agent.logger)
 
-        middleware = before_user_middlewares
-        middleware.extend(agent.middleware or [])
-        middleware.extend(after_user_middlewares)
+        self._agent_middleware: list[AgentMiddleware] = []
```
szykol, line 261: I'm not sure I like the fact that we're handling this in langchain.py now. I understand we also have some LC middleware and we have control over both SDK and LC middlewares here, but in my mind we should construct the SDK middleware in the SDK layer (like it was done before), not here aka this should belong to the SDK code, not the backend layer.

In case we wanted to switch to a different backend in the future we wouldn't need to move out this piece of code as well.

mateusz834: This is what I have partially done in the draft of this PR, but it didn't go well since the token limit to work properly needs access to the list of tools available, which in turn requires the use of LC mdidleware (see the next PR #770).  And for that reason i don't think we can abstract that out easily.

As an bit off-topic: Personally I think we should worry about abstraction for re-use between different backends only when these start to arise (i.e. we have 2nd backend), since a copy of these middlewares is not the worst thing.

splunklib/ai/README.md
mateusz834: Good catch.

splunklib/ai/engines/langchain.py
szykol: I agree with the sentiment, I think we already have a couple of things that we decided to leave for the future. However we already had those middlewares defined in the SDK layer, and we're moving it into backend layer.

If this fixes other issues let's keep it that way for now and figure out how we can move it back again in the future.

szykol (approved):

PR #782, opened 2026-05-12
Make AppInspect actually fail on error

Ickerday:


mateusz834 (approved):

.github/actions/run-appinspect/action.yml
```diff
@@ -19,4 +19,11 @@ runs:
...
       shell: bash
-      run: uvx splunk-appinspect inspect ./tests/system/test_apps/generating_app/mock_app.tgz --included-tags cloud
+      run: |
+        # AppInspect doesn't follow UNIX conventions so we swallow all non-zero
+        # exit codes unless its 101, the de facto error code
+        # https://dev.splunk.com/enterprise/reference/appinspect/appinspectcliref#inspect
+        uvx splunk-appinspect inspect \
+          ./tests/system/test_apps/generating_app/mock_app.tgz \
+          --included-tags cloud --ci || exit_code=$?
+        [ "${exit_code:-0}" -ne 101 ] || exit 1
```
szykol: curious whether we should handle the generic exit code = 1 as well here. Unless appinspect **never** returns it - just so the job doesn't fail quietly

mateusz834: I have a hard time understanding what this even does.
I would like to know what error coudes would cause the CI to fail and which not.

.github/actions/run-appinspect/action.yml
```diff
@@ -19,4 +19,11 @@ runs:
         tar -czf mock_app.tgz --exclude="__pycache__" bin default metadata
     - name: Validate the mock app with AppInspect
       shell: bash
-      run: uvx splunk-appinspect inspect ./tests/system/test_apps/generating_app/mock_app.tgz --included-tags cloud
+      run: |
+        # AppInspect doesn't follow UNIX conventions so we swallow all non-zero
+        # exit codes unless its 101, the de facto error code
```
mateusz834: When 101 is returned?

Ickerday: https://dev.splunk.com/enterprise/reference/appinspect/appinspectcliref#inspect
```
Exit code:
1 if the packaging standards checks failed;
2 if the validation encountered an exception;
3 if any check returned error result;
0 otherwise.

Additional exit codes if --ci option is provided:
101 if any check reported a failure
104 if any check reported a future_failure
103 if any check reported a warning
```

.github/actions/run-appinspect/action.yml
Ickerday: Yeah redid it now

Ickerday: Yeah remade it into a switch case with comments now

.github/actions/run-appinspect/action.yml
```diff
@@ -19,4 +19,15 @@ runs:
         tar -czf mock_app.tgz --exclude="__pycache__" bin default metadata
     - name: Validate the mock app with AppInspect
       shell: bash
-      run: uvx splunk-appinspect inspect ./tests/system/test_apps/generating_app/mock_app.tgz --included-tags cloud
+      run: |
+        # AppInspect doesn't follow UNIX conventions, gotta handle their exit codes explicitly
```
mateusz834, line 23: I don't see in what way it does not follow.

mateusz834: Not blocking of course.

PR #789, opened 2026-05-14
Update docker-compose configs

Ickerday:


Dockerfile
```diff
@@ -3,16 +3,16 @@ FROM splunk/splunk:${SPLUNK_VERSION}
 
 USER root
 
-# Copy splunk-mcp-server.tgz, we need to copy entire sdk since
+# Copy splunk-mcp-server.tgz, we need to copy entire SDK since
 # splunk-mcp-server.tgz might not exist and we don't want to fail in such case.
 RUN mkdir /tmp/sdk
+# TODO: Despite the comment, I don't get why we need to copy the entire SDK
```
Ickerday: @mateusz834 pls help :oooo

mateusz834: Since you can't dynamically (present or not present) copy a file in Dockerfile. If it didn't exist then it would have failed.

Dockerfile
```diff
@@ -24,8 +24,7 @@ USER splunk
 
 WORKDIR /tmp/sdk
 
-RUN /opt/splunk/bin/python3.13 -m venv .venv
-RUN /bin/bash -c "source .venv/bin/activate && LD_LIBRARY_PATH=/opt/splunk/lib python -m pip install '.[openai]' --target=/splunklib-deps"
+RUN /bin/bash -c "LD_LIBRARY_PATH=/opt/splunk/lib /opt/splunk/bin/python3.13 -m pip install '.[openai]' --target=/splunklib-deps"
```
mateusz834, line 26: Is this change the fix of the CI? Why this happens?

Python 3.13 was bundled already in previous version of splunk so i don't get the reason here.

mateusz834: (I think this change is fine, but just curious).

mateusz834 (approved):

PR #791, opened 2026-05-18
Move .coveragerc to pyproject.toml

Ickerday:


pyproject.toml
```diff
@@ -79,6 +79,8 @@ packages = [
 markers = ["app: requires sdk-app-collection", "smoke: essential smoke tests"]
 junit_family = "xunit2"
 
+[tool.coverage.run]
+omit = ["tests/*"]
```
mateusz834: Do we even use this coverage for anything?

mateusz834 (approved):

pyproject.toml
Ickerday: I don't think we ever reported it, but maybe it's a good idea to at least have a README badge or sth. Something to think about in the future. 😅

PR #795, opened 2026-05-20
Exclude **/bin/lib from basedpyright

Ickerday:


mateusz834:
I don't have any files in such folders, could you go over the reasoning of this? What files are stored there, and how a developer of this SDK can end up in such situation??

Ickerday:
@mateusz834 This is a safeguard against someone (or his agent) mistakenly committing any deps they vendor in an app when debugging or whatever else. We use this repo for miscellaneous, arbitrary purposes so it'd skip us some hassle when reviewing.

PR #796, opened 2026-05-21
fix: stream records lazily in write_records and _execute_chunk_v2

### LINKED ISSUE #687: Custom command have high CPU load / RAM usage
**Describe the bug**

I'm tagging this as a bug, but it's mainly a topic for discussion.

We have several custom commands in our environment and we've noticed a few issues:

- Significant CPU and RAM usage
- Preview mode not working

After profiling, we've noticed that our command takes only a few seconds to execute, but the Splunk search takes 5-10 minutes (on average) to complete.

We have observed that this occurs mainly on an SH cluster, and not on a standalone system (but to be confirmed).

**To Reproduce**

A little complicated to reproduce, but if you have a custom command that processes more than 150,000 Python objects (see 1 million for the test), you may encounter this problem, mainly on SH Cluster

**Expected behavior**

Fast search time, with or without Preview, affordable CPU/RAM consumption

**Logs or Screenshots**

For example, our custom command processed 75k events, our command took 3 seconds (calculated before and after the yield), but Splunk measured a Python script execution time of 460 seconds:

Image

**Splunk (please complete the following information):**

- Version: 9.3.1
- OS: RedHat 9
- Deployment: Search Head Cluster

- Version: 2.1.1
- Language Runtime Version: 3.9
- OS: RedHat 9

**Additional context**

Our question is, why the Splunk-SDK do a `list(records)` instead of send data by batch to the output?

https://github.com/splunk/splunk-sdk-python/blob/develop/splunklib/searchcommands/internals.py#L554 

We believe that this is the source of the problem. Instead of sending the data in batches to the standard output for Splunkd, Python loads all processed objects into memory, so the yield is no longer relevant from our custom command.

Seeing this, it now seems logical to us that preview mode does not work with custom commands and use lot of memory.

What do you think?

shivamsharma2106:
Fixes #687

### DIFF
--- a/splunklib/searchcommands/generating_command.py
+++ b/splunklib/searchcommands/generating_command.py
@@ -208,15 +208,11 @@ class GeneratingCommand(SearchCommand):
 
     def _execute_chunk_v2(self, process, chunk):
         count = 0
-        records = []
         for row in process:
-            records.append(row)
+            self._record_writer.write_record(row)
             count += 1
             if count == self._record_writer._maxresultrows:
                 break
-
-        for row in records:
-            self._record_writer.write_record(row)
 
         if count == self._record_writer._maxresultrows:
             self._finished = False
--- a/splunklib/searchcommands/internals.py
+++ b/splunklib/searchcommands/internals.py
@@ -527,7 +527,8 @@ class RecordWriter:
 
     def write_records(self, records):
         self._ensure_validity()
-        records = [] if records is NotImplemented else list(records)
+        if records is NotImplemented:
+            return
         write_record = self._write_record
         for record in records:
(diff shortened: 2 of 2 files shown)


mateusz834:
Hi, could you rebase this change on top of `develop`, since we had CI issues that are now fixed.

shivamsharma2106:
@mateusz834 Rebasing complete — branch is now on top of latest develop.

mateusz834:
I don't see it being rebased.

shivamsharma2106:
> I don't see it being rebased.

Can you check now please?

Ickerday:
Moving the discussion to #800

PR #825, opened 2026-07-07
Disallow logging.conf from elsewhere than the app folder

Ickerday:


splunklib/searchcommands/environment.py
mateusz834: Is that even correct? What if the CSC python file is inside `bin/foo/bar/csc.py` i don't think this will find the app dir correct, right?

There is an assumption that csc files are in `bin/csc.py`.

mateusz834, line 81: Shouldn't we update this branch too? It will hit when filename is a relative path.

mateusz834: I mean you could have a filename = `../../../outside.conf` and escape an app dir.

mateusz834: Ah, nevermind this will be detected in the branch you added too, since the `filename` var gets updated.

tests/unit/searchcommands/test_builtin_options.py
```diff
@@ -137,6 +137,30 @@ def test_logging_configuration(self):
...
+        except BaseException as e:
+            pytest.fail(f"Expected ValueError, but {type(e)} was raised")
+        else:
+            pytest.fail(
+                f"Expected ValueError, but logging_configuration={command.logging_configuration}"
+            )
+
+        # logging_configuration raises a value error when a relative path traverses outside the app directory (RCE guard)
+        try:
+            command.logging_configuration = os.path.join("..", "..", "..", "__init__.py")
```
mateusz834, line 163: please add an additional test with an absolute path, one that points to an app dir (success), and the other one that points to a different place (failure).

tests/unit/searchcommands/test_builtin_options.py
```diff
@@ -137,6 +137,30 @@ def test_logging_configuration(self):
                 f"Expected ValueError, but logging_configuration={command.logging_configuration}"
             )
 
+        # logging_configuration raises a value error for an existing absolute path outside the app directory (RCE guard)
+        try:
+            command.logging_configuration = os.path.join(package_directory, "__init__.py")
```
mateusz834: TBH I don't understand what this even tests.

`package_directory` is the location of some file in the SDK.

Does that test even make sense, if run outside of splunk? Since that is what we do here.

splunklib/searchcommands/environment.py
```diff
@@ -97,6 +97,12 @@ def configure_logging(logger_name, filename=None):
         global _current_logging_configuration_file
         filename = path.realpath(filename)
 
+        app_root_real = path.realpath(app_root)
+        if path.commonpath([filename, app_root_real]) != app_root_real:  # pyright: ignore[reportUnknownArgumentType]
```
mateusz834, line 101: If `filename` is relative here, what will happen? Does `realpath` always return the absolute path?

Docs does not seem to say anything about that:

```
def realpath(filename, *, strict=False):
    """Return the canonical path of the specified filename, eliminating any
symbolic links encountered in the path."""
```

It only follows symbolic links.

Ickerday: Seems like it does: https://github.com/python/cpython/blob/v3.13.13/Lib/posixpath.py#L388-L424

tests/unit/searchcommands/test_environment.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@pytest.mark.parametrize(
+    ("app_file_parts", "expected_app_root_parts"),
+    [
+        (("apps", "my_app", "bin", "command.py"), ("apps", "my_app")),
+        (("apps", "my_app", "bin", "linux_x86_64", "command.py"), ("apps", "my_app")),
+        (("apps", "my_app", "bin", "foo", "bar", "command.py"), ("apps", "my_app")),
+        (("some", "other", "layout", "command.py"), ("some", "other")),
+    ],
```
mateusz834, line 35: `(("apps", "my_app", "bin", "foo", "bin", "command.py"), ("apps", "my_app")),`

I think this would have failed.

`_find_app_root` should stop at `$SPLUNK_HOME/etc/apps/<name>/bin`.

splunklib/searchcommands/environment.py
```diff
@@ -114,9 +120,22 @@ def configure_logging(logger_name, filename=None):
...
 
+
+def _find_app_root(app_file: str) -> str:
+    """Return the app root directory for a search command script."""
+    directory: str = path.abspath(path.dirname(app_file))
+    while True:
+        parent, name = path.split(directory)
+        if name == "bin":
+            return parent
+        if parent == directory:
```
mateusz834: I don't understand this branch, could you add a comment, when it could hit?

mateusz834 (approved):

